PR #18, opened 2018-01-12
adding header and footer component

serenaf:
- Added navigation-header and navigation-footer component, which can be easily replaced later on by the proper components
- Fixed leaked state error with computed.sort, see https://github.com/emberjs/ember.js/pull/16084/files

cc @mansona

### DIFF
--- a/app/controllers/show.js
+++ b/app/controllers/show.js
@@ -2,6 +2,10 @@ import Controller from '@ember/controller';
 import EmberObject, { computed } from '@ember/object';
 
 export default Controller.extend({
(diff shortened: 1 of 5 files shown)


app/controllers/show.js
```diff
@@ -16,6 +16,6 @@ export default Controller.extend({
     });
     return result;
   }),
-  sortDefinition: ['since'],
+  sortDefinition:  Object.freeze(['since']),
```
mansona: why are you doing this?

PR #22, opened 2018-01-15
fixing some styling

serenaf:
First step in making the styling look closer to the current website. This PR contains the following changes:

- removed the container div of the navigation-footer and set the className binding of component to ```container```
- Ensure general structure of deprecations looks like website (renaming of headers)
- Cleanup and Accessibility work, copied over from website 

Current screenshot see below :) We are getting closer 🎉 

screen shot 2018-01-15 at 22 57 08


mansona:
you don't need to keep backup.css, if you delete it the diff will be way cleaner 

If we need it we can get it back from git 😂

Issue #26, opened 2018-01-21
make sure deprecations-app is using same header and footer as guides-app


serenaf:
https://github.com/ember-learn/guides-app/pull/20/files

jenweber:
PR https://github.com/ember-learn/deprecation-app/pull/35

serenaf:
Thanks @jenweber ❤️

PR #35, opened 2018-01-26
sync up header/footer with latest design work #26

jenweber:
Addresses #26. I can't tell for sure how this looks because the content is empty when I run it... Can you let me know how to run this locally, @serenaf ? 

Here's what I see on `ember serve`:
screen shot 2018-01-26 at 12 04 59 am


serenaf:
@jenweber Thanks a million Jen 🎉 Sorry i have not implemented an index page, so currently you need to go to ```http://localhost:4200/ember/v2.x``` to see data

serenaf:
Looks great with content as well ❤️ 
screen shot 2018-01-26 at 09 03 23

PR #38, opened 2018-01-28
supporting more content

serenaf:
This PR allows us to support multiple contents, such as ember-data, ember-cli and the various versions of ember

Please note that i only added an example deprecation for this Pr for each of the different versions. Will follow up with adding the actual content, once all these are supported. 
cc @mansona happy to chat through the changes and make improvements. Also i have no idea how the netlify setup works, its failing right now :(

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -4,20 +4,26 @@ const EmberApp = require('ember-cli/lib/broccoli/ember-app');
 const StaticSiteJson = require('broccoli-static-site-json');
 const BroccoliMergeTrees = require('broccoli-merge-trees');
 
-const jsonTree = new StaticSiteJson(`content/ember/v2`, {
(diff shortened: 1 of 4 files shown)


ember-cli-build.js
```diff
@@ -4,20 +4,26 @@ const EmberApp = require('ember-cli/lib/broccoli/ember-app');
 const StaticSiteJson = require('broccoli-static-site-json');
 const BroccoliMergeTrees = require('broccoli-merge-trees');
 
-const jsonTree = new StaticSiteJson(`content/ember/v2`, {
+const contentTypes = ['ember/v1', 'ember/v2', 'ember-data/v2', 'ember-cli/v2'];
```
mansona: I would change the name of this from `contentTypes` to `contentFolders`, it's clearer as to what it is (i.e. it doesn't actually change the type)

Issue #15, opened 2018-01-10
Check whether specific sorting is needed for toc


serenaf:
@locks is there a specific ordering you want if a specific version has multiple deprecations?

locks:
@serenaf assume sort by id for a first version, if need-be we can add a manual index and change the sorting logic!

jaredgalanis:
@serenaf @locks after bringing over the ember data files I'm realizing we need to revise the sorting because currently the CP uses the full versioning to sort (both major and minor). The issue with this is 2.10 will be listed before 2.3 because of the digits involved (i.e. 2.3 translates to 2.30 which is > 2.10). I have an idea for the sorting that focuses just on the minor version, which should sort things properly. I can open a PR on this in the next few days.

acorncom:
@jaredgalanis there should be code for that type of sort in the api docs app too ...

jaredgalanis:
Thanks @acorncom, I see that now. The `semver-compare` library is used there in the `getLastVersion` util, I'll be probably make use of `semver-compare` and borrow where needed from `getLastVersion`.

mansona:
fixed in https://github.com/ember-learn/deprecation-app/pull/54

PR #54, opened 2018-02-11
Adds server-compare and browserify, and updates sorting of grouped results accordingly.

jaredgalanis:
The sorting of the table of contents was out of order because of the way the semantic versioning numbering occurs (e.g. a normal sort treats 2.2 > 2.11). 

Using the semver-compare package (and browserify to import) which are used in the api-docs this sorting is now done properly.

The Upcoming Features group is moved to the end of the ordering scheme. 

Partially addresses #15, that issue still needs the grouped deprecations to be sorted.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "broccoli-asset-rev": "^2.4.5",
     "broccoli-static-site-json": "^1.3.0",
     "ember-ajax": "^3.0.0",
+    "ember-browserify": "^1.2.1",
     "ember-cli": "github:stonecircle/ember-cli#feature/merge-tree",
     "ember-cli-app-version": "^3.0.0",
     "ember-cli-babel": "^6.6.0",
@@ -49,7 +50,8 @@
     "ember-welcome-page": "^3.0.0",
     "eslint-plugin-ember": "^5.0.0",
     "loader.js": "^4.2.3",
-    "prember": "github:stonecircle/prember#feature/static-files"
+    "prember": "github:stonecircle/prember#feature/static-files",
+    "semver-compare": "^1.0.0"
   },
   "engines": {
     "node": "^4.5 || 6.* || >= 7.*"
--- a/package.json
+++ b/package.json
@@ -21,7 +21,6 @@
     "broccoli-asset-rev": "^2.4.5",
     "broccoli-static-site-json": "^1.3.0",
     "ember-ajax": "^3.0.0",
-    "ember-browserify": "^1.2.1",
     "ember-cli": "github:stonecircle/ember-cli#feature/merge-tree",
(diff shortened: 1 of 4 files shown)


rwjblue:
Generally speaking, we should try to avoid browserify. In this case the underlying implementation of semver-compare is _very_ small:

```
function cmp (a, b) {
    var pa = a.split('.');
    var pb = b.split('.');
    for (var i = 0; i < 3; i++) {
        var na = Number(pa[i]);
        var nb = Number(pb[i]);
        if (na > nb) return 1;
        if (nb > na) return -1;
        if (!isNaN(na) && isNaN(nb)) return 1;
        if (isNaN(na) && !isNaN(nb)) return -1;
    }
    return 0;
};
```

Let’s either inline it or use https://github.com/rwjblue/ember-cli-cjs-transform to app.import it...

jaredgalanis:
Thanks @rwjblue! Updating to ember-cli-cjs-transform now.

package.json
```diff
@@ -21,9 +21,11 @@
     "broccoli-asset-rev": "^2.4.5",
     "broccoli-static-site-json": "^1.3.0",
     "ember-ajax": "^3.0.0",
+    "ember-browserify": "^1.2.1",
```
rwjblue: Probably don’t need this any longer

mansona:
An alternative is to just app.import it directly like we do in the Guides App: https://github.com/ember-learn/guides-app/blob/master/package.json#L24 https://github.com/ember-learn/guides-app/blob/master/ember-cli-build.js#L72 where it becomes a global. 

@rwjblue is ember-cli-cjs-transform the **recommended** way to include this stuff? If so then I will probably convert the Guides App over too 👍

package.json
jaredgalanis: Ack! Yeah, sorry I missed that. I only ran npm uninstall.

jaredgalanis: Ack! Yeah, sorry I missed that. I just ran npm uninstall.

mansona:
@rwjblue I'm merging this but I would still love an answer re: ember-cli-cjs-transform so I can unify between here and guides-app if necessary 👍

rwjblue:
The transform is only needed because semver-compare is in CJS format. If you app imported something that was already using either UMD or globals, it would work without any special transformation. 

tldr; using app.import is good, and sometimes you need a transformation to deal with inconsistencies in the formatting of the incoming file.

mansona:
@rwjblue so to summarise: if we're using app.import() in the guides and it's working as expected it's not currently recommended that we also use ember-cli-cjs-transform?

rwjblue:
Yes, only use a special transform if you need it. 😃

PR #83, opened 2018-02-20
WIP - not for merge yet - has some questions / comments within

amyrlam:
@serenaf Doing checks for https://github.com/ember-learn/deprecation-app/issues/69, Ember v1, see https://www.emberjs.com/deprecations/v1.x/

//

WIP / not for merge yet - sorry this isn't the cleanest commit history, more was familiarizing myself with the repo and comparing to https://www.emberjs.com/deprecations/v1.x/. I have some questions inline too which I'll try to highlight here. Figured would be best to see if this is on the right track of what you are looking for...

//

Why does the until and id at the top of the .md file (in the dotted lines) show up inline on an ember serve? Such as:
image

Seems like titles at the top of the .md file (in the dotted lines) should not have a single tick?
image

Get rid of single tick formatting in links?
image

Issue #69, opened 2018-02-14
ensure all deprecation content is up-to-date


serenaf:
- [x] Ember v1, see https://www.emberjs.com/deprecations/v1.x/
- [x] Ember v2, see https://www.emberjs.com/deprecations/v2.x/
- [x] Ember v3, see https://www.emberjs.com/deprecations/v3.x/
- [x] Ember-Data v2, see https://www.emberjs.com/deprecations/ember-data/v2.x/
- [x] Ember CLI v2, see https://www.emberjs.com/deprecations/ember-cli/v2.x/

jenweber:
I think https://github.com/emberjs/website/pull/3099 also needs to be added

jenweber:
https://github.com/emberjs/website/pull/2995 too

amyrlam:
Took a stab at Ember v1. Can push future edits in a cleaner commit history...just wanted to get this up there to see if this is on the right track? https://github.com/ember-learn/deprecation-app/pull/83

amyrlam:
Made a few PR's to address v1. The only outstanding thing (I think?) on v1 is to add the missing warning section below. Will try to get to it tmrw. 

![image](https://user-images.githubusercontent.com/1372946/36465966-987ff894-168c-11e8-9932-2b63933c0b61.png)

serenaf:
Closing this for now, lets reopen specific issues for the missing content if there is any

PR #86, opened 2018-02-21
v1 deprecation tweaks

amyrlam:
Part of https://github.com/ember-learn/deprecation-app/issues/69

@serenaf Ready for review. There are a couple things missing from my first WIP pass https://github.com/ember-learn/deprecation-app/pull/83 - such as missing content - that I will incorporate in a separate PR.

### DIFF
--- a/content/ember/v1/non-standard-ways.md
+++ b/content/ember/v1/non-standard-ways.md
@@ -1,6 +1,6 @@
 ---
 id: non-standard-ways-of-calling-code-ember-set-code-and-code-ember-get-code
-title: Non-Standard Ways of Calling `Ember.set` and `Ember.get`
+title: Non-Standard Ways of Calling Ember.set and Ember.get
 until: ''
 since: '1.13'
 ---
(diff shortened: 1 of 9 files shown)


content/ember/v1/non-standard-ways.md
```diff
@@ -1,6 +1,6 @@
 ---
 id: non-standard-ways-of-calling-code-ember-set-code-and-code-ember-get-code
-title: Non-Standard Ways of Calling `Ember.set` and `Ember.get`
```
amyrlam: re: https://embercommunity.slack.com/archives/C04KG57CF/p1519239524000073 

I guess maybe we want to clean up the CSS of this instead to display backtick correctly. Because the `id` has to keep `code` in it. Sorry, I found this confusing and was just discussing `id`'s and `title`'s on Slack...

amyrlam: reverted here https://github.com/ember-learn/deprecation-app/pull/96

PR #87, opened 2018-02-21
Add missing section

amyrlam:
Part of https://github.com/ember-learn/deprecation-app/issues/69 cc @serenaf 

`Ember.ReduceComputed / Ember.ArrayComputed` from https://www.emberjs.com/deprecations/v1.x/#toc_ember-reducecomputed-ember-arraycomputed was missing.

Could be ok to merge as-is, but have questions if you care about these nits:

I set the `id`'s to be the same as the current website. But this makes `Ember.ReduceComputed / Ember.ArrayComputed` appear below `Ember.ReduceComputedProperty / Ember.ArrayComputedProperty`. 

Does the `id` naming matter? 

Where can I learn more about how the `id`'s are sorted? (Still getting familiar with the repo!)

### DIFF
--- a/content/ember/v1/ember-reduce-computed-property.md
+++ b/content/ember/v1/ember-reduce-computed-property.md
@@ -1,5 +1,5 @@
 ---
-id: ember-reducecomputed-ember-arraycomputed
+id: ember-reducecomputedproperty-ember-arraycomputedproperty
 title: Ember.ReduceComputedProperty / Ember.ArrayComputedProperty
 until: ''
 since: '1.13'
--- a/content/ember/v1/ember-reduce-computed.md
+++ b/content/ember/v1/ember-reduce-computed.md
@@ -1,0 +1,12 @@
+---
+id: ember-reducecomputed-ember-arraycomputed
+title: Ember.ReduceComputed / Ember.ArrayComputed
+until: ''
+since: '1.13'
+---
+
+These two little monsters served us well for a long time. They provided semantics that allowed to perform in-place modifications of an existing array instead of replacing the entire array like the regular `map/reduce` methods in javascript do, which was necessary to avoid expensive repaintings when rendering lists with the `{{each}}` helper.
+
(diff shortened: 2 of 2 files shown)


content/ember/v1/ember-reduce-computed-property.md
```diff
@@ -1,5 +1,5 @@
 ---
-id: ember-reducecomputed-ember-arraycomputed
+id: ember-reducecomputedproperty-ember-arraycomputedproperty
```
mansona: I'm not sure we can actually change the ids on anything because they are linked to by ember-cli console.log warnings. 

I'm not sure but it should be checked before we do update any ids.

amyrlam: Thanks for the explanation! It seemed like this `id` was wrong based on the old app? But I wasn't sure if the current emberjs.com website url and `id` were necessarily linked?

re: https://www.emberjs.com/deprecations/v1.x/#toc_ember-reducecomputed-ember-arraycomputed and 
https://www.emberjs.com/deprecations/v1.x/#toc_ember-reducecomputedproperty-ember-arraycomputedproperty

mansona: so there are some deployment issues that I'm trying to sort out on the new deprecations app right now but essentially you should be able to see what the links look like you can prefix the # link with `https://deprecations-app-prod.herokuapp.com/ember/v1.x/` (it should work the first time and then break on refresh but that's expected right now 😖) 

so you will notice that if you go to https://deprecations-app-prod.herokuapp.com/ember/v1.x/#toc_ember-reducecomputed-ember-arraycomputed it doesn't scroll down the page but going to https://deprecations-app-prod.herokuapp.com/ember/v1.x/#toc_ember-reducecomputedproperty-ember-arraycomputedproperty does ... or at least it should... blast it! 😖 

So I'm not exactly sure exactly what the **right** way is... but the thing I *know* is that we need to maintain inbound links. It's potentially a good idea to raise the issue of QA right now to make sure we are in fact supporting the same links 🤔

amyrlam:
I'm gonna close this since it's stale, but my goal is to get more involved on this repo and help @mansona

Issue #92, opened 2018-02-21
Serena: Id-checking


serenaf:
This is a WIP issue for me to check for all the ids:
 Version 1:
- [x] #toc_observing-container-views-like-arrays
- [x] #toc_ember-deferredmixin-and-ember-deferred
- [x] #toc_code-then-code-on-ember-application
- [x] #toc_global-lookup-of-views
- [x] #toc_hash-location-paths-without-leading-slashes
- [x] #toc_more-consistent-handlebars-scope
- [x] #toc_objectcontroller
- [x] #toc_access-to-instances-in-initializers
- [x] #toc_code-in-code-syntax-for-code-each-code
- [x] #toc_code-as-code-syntax-for-code-with-code
- [x] #toc_computed-properties-with-a-shared-getter-and-setter
- [x] #toc_ember-create
- [x] #toc_ember-keys
- [x] #toc_ember-oneway
- [x] #toc_ember-view
- [x] #toc_ember-containerview
- [x] #toc_ember-collectionview
- [x] #toc_set-code-positionalparams-code-as-a-static-property-on-the-class
- [x] #toc_ember-reducecomputed-ember-arraycomputed
- [x] #toc_modifying-a-property-within-code-didinsertelement-code
- [x] #toc_beforeobserver
- [x] #toc_ember-immediateobserver
- [x] #toc_arraycontroller
- [x] #toc_sortablemixin
- [x] #toc_renderbuffer
- [x] #toc_using-code-guid-code-and-code-item-code-as-key-in-code-each-code
- [x] #toc_controller-needs
- [x] #toc_block-and-multi-argument-unbound-helper
- [x] #toc_block-and-multi-argument-unbound-helper
- [x] #toc_ember-freezable
- [x] #toc_ember-enumerableutils
- [x] #toc_copyable-frozencopy
- [x] #toc_using-code-code-for-namespace-in-the-code-render-code-helper
- [x] #toc_non-standard-ways-of-calling-code-ember-set-code-and-code-ember-get-code
- [x] #toc_reversed-ember-observer-arguments
- [x] #toc_ember-createwithmixins
- [x] #toc_bind-attr
- [x] #toc_using-code-this-get-template-code
- [x] #toc_ember-computed-any
- [x] #toc_view-and-controller-options-on-the-code-each-code-helper
- [x] #toc_using-the-code-with-code-helper-with-the-code-controller-code-option
- [x] #toc_handlebars-htmlbars-helpers
[...]

PR #93, opened 2018-02-21
fixing all the ids

serenaf:
Handselected Ids from the website, see https://github.com/ember-learn/deprecation-app/issues/92 for reference

cc @amyrlam @mansona


serenaf:
@mansona Chris, do you have any idea why there is always the id added to the id string? 😢 see https://deprecations-app-staging-pr-93.herokuapp.com/deprecations/ember/v2.x#toc_id-ember-component-didinitattrs as an example

Issue #95, opened 2018-02-23
Fix code (single ``) styling


amyrlam:
Since the `id` for this title is "locked" as legacy `id: non-standard-ways-of-calling-code-ember-set-code-and-code-ember-get-code` (aka including the backticks written out as `code`), we should preserve the backticks in the `title`. Relates to: https://github.com/ember-learn/deprecation-app/pull/96.

(1) We should fix the styling of backtick code in the left hand side table of contents.

![image](https://user-images.githubusercontent.com/1372946/36582525-bda4c7f8-1826-11e8-931a-cc32b38d0c0b.png)

(2) We should fix the styling of backtick code in the title in-line.

![image](https://user-images.githubusercontent.com/1372946/36582518-b7abca22-1826-11e8-9f01-78ae00f98be1.png)

jaredgalanis:
Will be addressed by https://github.com/ember-learn/deprecation-app/pull/793

jaredgalanis:
Closed by #793

PR #96, opened 2018-02-23
Put `code` backtick back in titles

amyrlam:
* Revert "Cleanup titles". This reverts commit c3e7677 from https://github.com/ember-learn/deprecation-app/pull/86. We need to preserve the word `code` in the id, so it is confusing to remove the backtick from the title. Relates to https://github.com/ember-learn/deprecation-app/issues/95.

This keeps the `id` and `title` matching, example below:
```
id: overriding-code-render-code-when-extending-a-component-or-view
title: Overriding `render` When Extending a Component or View
```

* Fix unnecessary period in `title` (did not match other titles without a period).


serenaf:
@amyrlam 🎉 thanks a million Amy :)

PR #99, opened 2018-02-24
Added instructions for add-on authors

lupestro:
Additional text from discussion with @rwjblue and @locks after feedback from add-on authors whose add-ons must support both Ember 2.x and 3.x.


serenaf:
@lupestro Thank you so much for this addition 🎉 
Apologies for the delay in reviewing. Will merge this now 🥇

PR #103, opened 2018-03-10
updating deprecation ID

mansona:
This is based on a link that @rwjblue shared: https://emberjs.com/deprecations/v3.x#toc_use-console-rather-than-ember-logger

If this is the right link (i.e. it's what's in ember-cli) then merge this 👍


mansona:
@rwjblue confirmed that he would update the link in the ember source 👍 #nothingtoseehere

PR #104, opened 2018-03-12
[WIP] making the html structure better

serenaf:
This PR is a work in progress in making the HTML structure of the deprecations app more sane :)

@MelSumner Could you have a look with me what should be done here? Right now it ads a border-bottom to the h1 element, because of below rule. Can i override this rule? 

screen shot 2018-03-12 at 10 07 29
screen shot 2018-03-12 at 10 07 22


MelSumner:
If there's a sidebar it will put a line under because that's consistent with other places. Did we merge this?

serenaf:
@MelSumner No sorry Mel, this was not merged. I will re-open a new PR with this change and some other small cleanups

PR #105, opened 2018-03-15
Add command to generate new deprecations

locks:
[edit] I am now ok with shipping this as is.

cc @serenaf

### DIFF
--- a/blueprints/deprecation/index.js
+++ b/blueprints/deprecation/index.js
@@ -1,0 +1,18 @@
+/* eslint-env node */
+module.exports = {
+  description: 'Generate a new deprecation file',
+
+  locals(options) {
+    // Return custom template variables here.
+    return {
+      name: extractDeprecationId(options.entity.name)
+    };
+  }
+};
+
+function extractDeprecationId(path) {
+  var pathParts = path.split('/')
+  var name = pathParts[pathParts.length - 1].split('.')[0]
+
+  return name;
+}
--- a/README.md
+++ b/README.md
@@ -4,7 +4,13 @@ This is the app that serves https://emberjs.com/deprecations/
 
 ## Adding new deprecations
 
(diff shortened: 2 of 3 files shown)


mansona:
@locks I **love** this 😍 I would be interested in figuring out the "right" way to do these options but also to surface that in the docs for the ember-cli. It could potentially be a good "tracer bullet" for the ember-cli docks improvements

locks:
@mansona I tried a couple approaches and `ember/v3/deprecate-ember-logger` seems like the easiest? 🤔

yarn.lock
```diff
@@ -2314,18 +2314,6 @@ core-js@^2.4.0, core-js@^2.4.1, core-js@^2.5.0:
   version "2.5.3"
   resolved "https://registry.yarnpkg.com/core-js/-/core-js-2.5.3.tgz#8acc38345824f16d8365b7c9b4259168e8ed603e"
 
-core-object@2.0.6:
```
Turbo87: these changes seem unrelated?

blueprints/deprecation/index.js
```diff
@@ -0,0 +1,18 @@
...
+    // Return custom template variables here.
+    return {
+      name: extractDeprecationId(options.entity.name)
+    };
+  }
+};
+
+function extractDeprecationId(path) {
+  var pathParts = path.split('/')
+  var name = pathParts[pathParts.length - 1].split('.')[0]
```
Turbo87: it's not quite obvious what this code extracts. since there are no tests for this it should at least have some comment with an example that explains the behavior.

yarn.lock
locks: ✅

blueprints/deprecation/index.js
locks: Can you write the test? 😅

PR #110, opened 2018-03-21
Mention ember-notify-property-change-polyfill

rondale-sc:


content/ember/v3/use-notify-prop-change.md
```diff
@@ -33,3 +33,7 @@ Ember.notifyPropertyChange(object, 'someProperty');
 doStuff(object);
 object.notifyPropertyChange('someProperty');
 ```
+
+If you are an addon author and need to support both you can use the polyfill
```
rwjblue: Slight tweak here: you should mention “both” of what...

rondale-sc: Updated to, hopefully, be more clear.

acorncom:
Thanks @rondale-sc, looks good.

Issue #118, opened 2018-04-17
Move `Ember.String` deprecation to v3


locks:
Someone—who totally isn't me,—delayed the `Ember.String` deprecation long enough that whenever it lands, it will be in the 3.x series.

We need to move the following two files from v2 to v3:
* https://github.com/ember-learn/deprecation-app/blob/master/content/ember/v2/ember-string-namespace.md
* https://github.com/ember-learn/deprecation-app/blob/master/content/ember/v2/ember-string-prototype-extensions.md

To do that you need to move the files from one folder to another, and then update the "until" to "4.0.0".

You don't need to worry about the content for now.

GavinJoyce:
Looks like this can be closed

PR #119, opened 2018-04-17
Move `Ember.String` deprecation to v3

albertpak:
Someone—who totally isn't me,—delayed the Ember.String deprecation long enough that whenever it lands, it will be in the 3.x series.

We need to move the following two files from v2 to v3:

https://github.com/ember-learn/deprecation-app/blob/master/content/ember/v2/ember-string-namespace.md
https://github.com/ember-learn/deprecation-app/blob/master/content/ember/v2/ember-string-prototype-extensions.md
To do that you need to move the files from one folder to another, and then update the "until" to "4.0.0".

You don't need to worry about the content for now.

Issue #118

### DIFF
--- a/content/ember/v3/ember-string-namespace.md
+++ b/content/ember/v3/ember-string-namespace.md
@@ -1,7 +1,7 @@
 ---
 id: ember-string.namespace
 title: Ember.String namespace
-until: '3.5.0'
+until: '4.0.0'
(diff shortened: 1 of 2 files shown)


serenaf:
Thanks so much @albertpak for your contribution 🎉

locks:
Done 😁

PR #126, opened 2018-05-27
[RFC322] Deprecate copy copyable

lupestro:
This is the deprecation notice that goes with the PR just merged for deprecating copy/copyable.


lupestro:
@jenweber Good catch! I cleaned up the wording. The `copy` method of `Copyable` is documented to do both shallow and deep copies and the implementation in `NativeArray` is a faithful implementation. I think the (fewer) words in the edit say that without raising a further unnecessary question to be answered.

acorncom:
Thanks @josemarluedke, it sure should 👍

PR #127, opened 2018-05-27
Add jQuery.Event deprecation

simonihmig:
As specified by https://github.com/emberjs/rfcs/blob/master/text/0294-optional-jquery.md#deprecation-guide

### DIFF
--- a/content/ember/v3/jquery-event.md
+++ b/content/ember/v3/jquery-event.md
@@ -1,0 +1,83 @@
+---
+id: jquery-event
+title: Use native events instead of jQuery.Event
+until: '4.0.0'
+since: '3.3'
+---
+
+As part of the effort to decouple Ember from jQuery, using event object APIs that are specific to `jQuery.Event` such as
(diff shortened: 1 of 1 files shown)


content/ember/v3/jquery-event.md
```diff
@@ -0,0 +1,83 @@
+---
+id: jquery-event
+title: Use native events instead of jQuery.Event
+until: '4.0.0'
+since: '3.3'
```
simonihmig: Not totally sure how this is supposed to be... the deprecation is in master now, so I expect this to land in 3.3. But as this has not been released yet, should this better be left out?

sivakumar-kailasam: We expect users to see this page only when they get the link to this deprecation on their browser console. 

So it should be fine to have it as 3.3 for now. If you have a quest style issue for this somewhere else, then it probably helps to add an item to update this if it doesn't make it into 3.3.

sivakumar-kailasam:
Leaving this open for @rwjblue/@locks to review

rwjblue:
Thank you!

PR #145, opened 2018-06-15
adding back in the warning for style attributes

mansona:
So this was missing from the conversion to the new Deprecations app but I have a feeling I know why.

This isn't a deprecation, it's a warning. This is now just being considered a deprecation because it will fix the inbound link in warning in the Ember code base: https://emberjs.com/deprecations/v1.x/#toc_binding-style-attributes

There is probably a better way to do this, by marking it as a warning and having a new group type for warnings or something but I think it's overkill considering this is literally the only "warning" in the entire system.

### DIFF
--- a/content/ember/v1/binding-style-attributes.md
+++ b/content/ember/v1/binding-style-attributes.md
@@ -1,0 +1,45 @@
+---
+id: binding-style-attributes
+title: Binding Style Attributes
+until: ''
+since: '1.11'
+---
+<a id="toc_warning-when-binding-style-attributes"></a>
+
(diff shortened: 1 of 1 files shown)


mansona:
It looks like the travis failure was due to a yarn issue: 

screen shot 2018-06-15 at 16 38 58

Looks like there was an error getting the `install.sh` from curl, looks like a temporary outage or something. In fact it's a little scary that it's doing that at all 😬 

I've restarted the tests

rwjblue:
Thank you for fixing!

PR #146, opened 2018-06-15
Add deprecation entry for Component#sendAction

cibernox:


content/ember/v3/send-action.md
```diff
@@ -0,0 +1,87 @@
+---
+id: ember-component.send-action
+title: Use closure actions instead of `sendAction`
+until: '4.0.0'
+since: '3.3'
+---
+
+In Ember 1.13 closure actions were introduced as a recommended replacement for `sendAction`.
+With `sendAction` the developer passes the name of of an action, and when `sendAction` is called Ember.js
```
sbatson5: Repeated word `of of and action`

content/ember/v3/send-action.md
```diff
@@ -0,0 +1,87 @@
...
+
+```js
+// my-component.js
+this.submit();
+```
+
+Note that with this approach the component MUST receive that `submit` property, while with `sendAction` if
+it didn't it would silently do nothing.
+
+If you dont want `submit` to be mandatory, you have to check for the presence of the action before calling it:
```
sbatson5: Missing apostrophe on `don't`

content/ember/v3/send-action.md
```diff
@@ -0,0 +1,87 @@
...
+}
+```
+
+This deprecation also affects the built-in `{{input}}` helper that used to allow action invocations like:
+
+```hbs
+{{input enter="handleEnter"}}
+```
+
+Since this used `sendAction` underneath is not also disallowed and must also be replaced by closure actions:
```
sbatson5: `is not also disallowed` is a bit confusing. Are you saying the use of `enter="handleEnter"` is disallowed?

content/ember/v3/send-action.md
```diff
@@ -0,0 +1,87 @@
+---
+id: ember-component.send-action
+title: Use closure actions instead of `sendAction`
+until: '4.0.0'
+since: '3.3'
```
rwjblue: The `since` here would be at least 3.4 (emberjs/ember.js#master represents 3.4 ATM)

content/ember/v3/send-action.md
rwjblue: You could also possibly suggest defaulting the value on the prototype (so that you invoke without a guard).

```js
export default Component.extend({
  submit() {} // <- empty function so that `this.submit()` always works
});
```

cibernox:
Updated!

PR #187, opened 2018-09-20
Add notes for how it deploys to the README

jenweber:
@serenaf can you take a look over this PR and also the README in general? If there are any other issues we should write up for Hacktoberfest or anything we can do to make it easier for new contributors, let me know.


serenaf:
Looks great Jen 🙌 I check the next weekend for some good issues in here (on holidays in Italy for a week😎)

PR #191, opened 2018-09-27
[WIP] Add the ability to filter deprecations by range

cafreeman:
# What this does
This PR adds a new `/filters` route, which exposes all of the deprecations as a filterable list.

Currently, you can only filter by version range (meaning that ember, ember-cli, and ember data are all lumped in together).

# New dependencies
This PR adds several new dependencies:
- `qunit-dom` & `ember-test-selectors` (both for acceptance testing)
- `semver` for comparing version range. This PR currently uses `semver-compare`, but its API is much more limited, and for the sake of filtering ranges in query params, `semver`'s more diverse API proved to be very useful
- `emberx-select` for the filter dropdowns

# Major changes
- New route, `/filters`. I am 100% *not* attached to this name at all, just had to start somewhere 😛 
- Moved the `contentFolders` list out of the `content-docs-generator` addon and into a standalone module so that it can be imported elsewhere.
- Exposed `contentFolders` list on `config` object so that it can be used to grab all of the deprecation messages in the `filters` route.

# Things worth noting

The current deprecation pages use `path` and `version` query params to request JSON from the `content` path. This works fine in those cases because each route in the deprecations app renders static chunks of deprecations (e.g. all the deprecations in ember data 2.x).

However, to filter across all sorts of ranges, I didn't see a way to do that via params sent in an ember data query, so instead I just load *all* of the deprecation warnings in the `/filters` route and then do the filtering via a set of computed properties in the controller.

# WIP status

This is 100% just a first-pass on the functionality that @locks and I discussed initially. Right now there's not much styling, and we could probably use some copy at the top of the page to explain what's going on. Additionally, the `index` route just contains a link that says `Filters page`, basically just so you can access it during dev. I'm not sure how we'll want to tie in the filters with the current deprecation categories.

For reference, here's the filters link

image

In addition, here's a screen shot of the (WIP) filters page:

image

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -24,7 +24,8 @@ module.exports = {
         'testem.js',
         'blueprints/*/index.js',
         'config/**/*.js',
-        'lib/*/index.js'
+        'lib/*/index.js',
+        'content-folder-list.js'
       ],
       parserOptions: {
         sourceType: 'script',
--- a/app/controllers/filters.js
+++ b/app/controllers/filters.js
@@ -1,0 +1,68 @@
+import Controller from '@ember/controller';
+import { computed } from '@ember/object';
+import semverCompare from 'semver-compare';
+
+export default Controller.extend({
+  queryParams: ['from', 'to'],
+
+  from: null,
+  to: null,
+
+  flattenedDeprecations: computed('model.[]', function() {
+    return this.model.reduce((acc, elem) => {
+      let rawArray = elem.toArray();
+      return acc.concat(rawArray);
+    }, []);
+  }),
+
+  versions: computed('flattenedDeprecations.@each.since', function() {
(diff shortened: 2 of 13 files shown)


sivakumar-kailasam:
This is a great start @cafreeman. Does this page have a table of contents like the individual versions page does?

cafreeman:
@sivakumar-kailasam It doesn't at the moment, but only because I don't really have a design for what all should/shouldn't go here (besides the basic filter functionality). I'm definitely open to suggestions!

jenweber:
Hello! Thanks for this work! Another contributor has updated this branch and work has continued over here: https://github.com/ember-learn/deprecation-app/pull/361 Now that Octane is out, we are revisiting some of these older PRs and hoping to get them shipped soon.

Issue #197, opened 2018-10-08
Copy search components from the guides-app to the deprecation-app and make sure the algolia search is configured with the deprecation-app index


mansona:
You can see the components in https://github.com/ember-learn/guides-app/pull/226

The index will need to be created & populated by https://github.com/ember-learn/deprecation-app/issues/196 so I wouldn't start this until that is done

PR #200, opened 2018-10-09
adding Algolia indexing config

### LINKED ISSUE #196: Add ember-cli-deploy-algolia to deprecations app and make sure it is indexed
Essentially we need to implement the same "release" pipeline for the deprecations app as we have for the Guides app

You can see some of what is needed here: https://github.com/ember-learn/guides-app/pull/226

We should run the index manually for now

It is probably best to wait for https://github.com/ember-learn/deprecation-app/issues/195 before implementing this

mansona:
This has already been run and is in the Algolia index named `ember-deprecations`

Fixes #196 

Once this is merged work can start on #197


mansona:
@sivakumar-kailasam I think adding ember-dotenv might be a bit overkill in this situation 🤔 we don't need it to run the app in any way and there is no need for the "failOnMissingKey" functionality.

This will just fail on `ember deploy production` if the env files are missing and anyone who is doing that will know & have the power to fix it

PR #208, opened 2018-10-19
Fix ID of "Getting the @each property" deprecation

deverstalmage:
The correct ID can be found here: https://github.com/emberjs/ember.js/blob/c2c8171301d2b1a0c499b4a55009ebf428f40e69/packages/%40ember/-internals/runtime/lib/mixins/array.js#L1159

### DIFF
--- a/content/ember/v3/getting-each.md
+++ b/content/ember/v3/getting-each.md
@@ -1,5 +1,5 @@
 ---
-id: getting-the-each-property
+id: ember-metal.getting-each
 title: Getting the @each property
 until: '3.5.0'
 since: '3.1'


mansona:
Thanks for the contribution @deverstalmage 

Might I ask why you're updating this? Did you try to follow a deprecation link and it didn't work? I'm just asking because we have to be quite careful with these IDs and I wanted to know what prompted you to want to change it so that we can verify that it's right 😄

deverstalmage:
@mansona yep, I was trying to figure out why silencing the deprecation wasn't working, and it was due to the wrong deprecation ID being reported. Once I switched to the `ember-metal.getting-each` ID it silenced the deprecation as I intended.

mansona:
@deverstalmage when you say "silencing the depreciation wasn't working" what do you mean exactly? It this using the ember-inspector? 

I'm sorry if I'm seeming a bit slow on this one, like I said we have to be very careful with these IDs and I want to know enough so I can go pull on a thread and figure out why this seems wrong (or is wrong)

deverstalmage:
Should have been more clear. 

My org uses a tool internally that allows you to specify a blacklist of deprecation warnings and will not forward these errors/warnings to our bug tracking software. The deprecations are specified using their deprecation IDs, and are compared as strings vs the deprecation ID reported in the deprecation warning output by Ember. 

The ID that is appearing in the deprecation message, `getting-the-each-property`, does not match the actual deprecation ID specified in the `ember-metal` package. You can view the ID in the `ember-metal` source here: https://github.com/emberjs/ember.js/blob/c2c8171301d2b1a0c499b4a55009ebf428f40e69/packages/%40ember/-internals/runtime/lib/mixins/array.js#L1159 As you will note, the ID in the source code is `ember-metal.getting-each`. The link that `ember-metal` specifies is https://www.emberjs.com/deprecations/v3.x/#toc_getting-the-each-property and the ID displayed on that page (which this PR is addressing) is `getting-the-each-property` and does not match the ID reported by `ember-metal`. I've included an image highlighting the differences (between the two docs that I believe are supposed to match up) below:

screen shot 2018-10-30 at 1 59 54 pm
screen shot 2018-10-30 at 1 59 26 pm

deverstalmage:
@mansona anything else I need to change in order to get this merged?

mansona:
@deverstalmage I've been digging into this issue for the past few days and I'm trying to replicate something... 

When I asked about where you are seeing these IDs and how it's affecting you, you said: 

> My org uses a tool internally that allows you to specify a blacklist of deprecation warnings

Is that a public tool? I'm trying to fix your issue completely and to do so I need to fully understand the problem space. Are you using the deprecations page on the website as a reference to look up the ID, and that's why it matters that the ID is correct on the page?

rwjblue:
@mansona I was thinking they meant ember-cli-deprecation-workflow

PR #266, opened 2019-01-23
Remove erraneous Route#willTransition and Route#didTransition deprecation guidance

### LINKED ISSUE #253: Potential issue with code snippet for router events deprecation in Ember 3.6 section
Hi.  I'm sorry if this is not the right place to post this but I have been having a discussion with a few members of the community about this deprectaion:

https://deprecations-app-prod.herokuapp.com/deprecations/v3.x/#toc_deprecate-router-events

In the deprecations guide there is a code example of how to replace the willTransition/didTransition hooks with routeWillChange/routeDidChange.  The example says that the new event listeners should be registered in the init() method but this results in the new events firing as you transition to AND away from a particular route. To mimic willTransition the routeWillChange event should only fire when transitioning away from the route. Likewise, in order to mimic didTransition the routeDidChange event should only fire as you enter a route. 

Another community member has suggested that the correct method that should be used to register the new listeners is activate() rather than init().  This does result in the correct behaviour where routeWillChange mimics willTransition etc.

How do we go about deciding if the code snippet should be modified? Do we need to contact the deprecation author for confirmation?

Thanks.

efx:
Fixes #253. See conversations in ember-dev-learning on discord:
https://discordapp.com/channels/480462759797063690/480777444203429888/536988540010102784

### DIFF
--- a/content/ember/v3/router-events.md
+++ b/content/ember/v3/router-events.md
@@ -9,64 +9,8 @@ Application-wide transition monitoring events belong on the Router service, not 
 
 In addition, they receive handlerInfos in their arguments, which are an undocumented internal implementation detail of router.js that doesn't belong in Ember's public API. Everything you can do with handlerInfos can be done with the `RouteInfo`.
 
-Given services are [singletons](https://en.wikipedia.org/wiki/Singleton_pattern), you may need to adjust some of the behavior of Route specific `willTransition` or `didTransition` logic.
 
-Below is how you would transition both the `Route` and `Router` usages of `willTransition` and `didTransition`.
-
-### Route
-
-From:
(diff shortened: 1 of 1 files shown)


efx:
[Rendered](https://github.com/efx/deprecation-app/blob/fix-253/content/ember/v3/router-events.md)

rwjblue:
@chadhietala - Can you review?

efx:
hold off on merging, I think there is some confusion around `willTransition` `didTransition` deprecations per route.

efx:
Ok, per the discussion in discord ember-js-dev, I think this is the correct deprecation guidance.

PR #276, opened 2019-02-06
Application controller router properties deprecation

locks:


jenweber:
Should we wait to merge this until 3.10 lands?

locks:
@jenweber up to you!

rwjblue:
No, folks testing betas/canaries also see these urls, they should be valid nearly as soon as the deprecation itself is landed.

tldr; we should land these PR's ASAP 😸

PR #302, opened 2019-03-07
Allow to be used as both a subdomain and a subfolder

### LINKED ISSUE #301: In-page anchors do not scroll correctly on a fresh load
If I want to share the URL of a particular deprecation like this one: https://deprecations-app-prod.herokuapp.com/deprecations/v3.x/#toc_component-manager-string-lookup

The page does not actually scroll to that place in the page making it very hard to find.

mansona:
This gives us the ability to run this app in both a subfolder (as it is currently) and as a domain directly. This allows us to have a small amount of time where the app is running in both places at the same time for the transition period (no down time)

This also fixes prember by downgrading ember (I will file a bug later to fix this)

And I added a test to make sure that prember is working 👍 

Fixes #301


mansona:
Also this is actually working now but there seems to be a second netlify app installed on the repo that is not part of the official Ember netlify team 🤔 

Do we know who owns https://app.netlify.com/sites/ember-deprecation-app?

.travis.yml
```diff
@@ -10,20 +10,19 @@ addons:
...
-  - export PATH=$HOME/.yarn/bin:$PATH
-
-install:
-  - yarn install --no-lockfile --non-interactive
+  - npm install -g npm@6
+  - npm config set spin false
 
 script:
-  - yarn lint:js
-  - yarn test
```
acorncom: I think I'm missing some context here, did we end up deciding we were going to shift **away** from Yarn? I'd thought our decision had been to stick **with** yarn as a team last I knew ...

acorncom:
> Do we know who owns https://app.netlify.com/sites/ember-deprecation-app?
@mansona looks like I might be the one with the ownership here, do we want to turn it off?

.travis.yml
jenweber: In general, we try to use yarn, however it appears that this project already uses npm (based on the lockfiles) so this doesn't appear to be a "change" to me. I added an approving review to move things along, but it would be good to have the rationale recorded here

mansona:
@acorncom yes please 🎉 you can just delete the netlify app and then things will be back to normal 👍

PR #361, opened 2019-06-04
Filterable deprecations

NullVoxPopuli:
Continuation of: https://github.com/ember-learn/deprecation-app/pull/191

stuff doesn't work, don't know why :man_shrugging:


MelSumner:
@NullVoxPopuli @mansona any thoughts on this one?

NullVoxPopuli:
I think filtering is useful, but I don't know how much of this implementation is salvageable. especially given that there are errors that idk if they are worth investigating. Maybe things from this attempt could be pulled in to a fresh branch though

jaredgalanis:
> I think filtering is useful, but I don't know how much of this implementation is salvageable. especially given that there are errors that idk if they are worth investigating. Maybe things from this attempt could be pulled in to a fresh branch though

The errors seem to be coming from `yarn install`. When I blow away `node_modules` and `npm install` I do not get the errors (I was also getting them with yarn).

PR #436, opened 2019-09-09
RFC-0522 Ember-Data Deprecate default serializers and adapters

pete-the-pete:
Adding deprecation documentation for RFC-0522

### DIFF
--- a/content/ember-data/v3/default-adapter.md
+++ b/content/ember-data/v3/default-adapter.md
@@ -19,7 +19,7 @@ create the file app/adapters/application.js with the following:
 ```
 
 
-If you were setting the adapter property previously to <adapter-name>, create the file app/adapters/application.js with the following:
+If you were setting the adapter property previously to `<adapter-name>`, create the file app/adapters/application.js with the following:
 
 ```js
     export { default } from './<adapter-name>';
--- a/content/ember-data/v3/default-adapter.md
+++ b/content/ember-data/v3/default-adapter.md
@@ -1,0 +1,10 @@
+---
+id: ember-data:default-adapter
+title: Default adapter usage
+until: '4.0.0'
+since: '3.12.0'
+---
+## deprecate store.defaultAdapter (-json-api) and the -json-api adapter fallback behavior
+A deprecation warning will be shown when the store's `defaultAdapter` is accessed.
+
+To remove this deprecation warnings, users should implement an application adapter.
--- a/content/ember-data/v3/default-adapter.md
+++ b/content/ember-data/v3/default-adapter.md
@@ -12,3 +12,5 @@ To remove this deprecation warnings, create an `app/adapters/application.js` wit
     import { JSONAPIAdapter } from '@ember-data/adapters';
 
     export default JSONAPIAdapter.extend();
+
+More information about custom adapters can be found on the [ember.js/guides](https://guides.emberjs.com/release/models/customizing-adapters/)
--- a/content/ember-data/v3/default-adapter.md
+++ b/content/ember-data/v3/default-adapter.md
@@ -14,12 +14,16 @@ If you were not setting this value previously, the following should be sufficien
 
 create the file app/adapters/application.js with the following:
 
-export { default } from '@ember-data/adapters/json-api';
+```js
+    export { default } from '@ember-data/adapters/json-api';
+```
 
 
 If you were setting the adapter property previously to <adapter-name>, create the file app/adapters/application.js with the following:
 
-export { default } from './<adapter-name>';
+```js
+    export { default } from './<adapter-name>';
+```
 
 
  More information about custom adapters can be found on the [ember.js/guides](https://guides.emberjs.com/release/models/customizing-adapters/) and in the [API DOCs](https://api.emberjs.com/ember-data/release/modules/@ember-data%2Fadapter)
--- a/content/ember-data/v3/default-adapter.md
+++ b/content/ember-data/v3/default-adapter.md
@@ -7,7 +7,7 @@ since: '3.12.0'
 ## deprecate store.defaultAdapter (-json-api) and the -json-api adapter fallback behavior
 A deprecation warning will be shown when the store's `defaultAdapter` is accessed.
 
-To remove this deprecation warnings, create an `app/adapters/application.js` with the following:
+To remove this deprecation warning, create an `app/adapters/application.js` with the following:
 
     import { JSONAPIAdapter } from '@ember-data/adapters';
 
--- a/content/ember-data/v3/default-adapter.md
+++ b/content/ember-data/v3/default-adapter.md
@@ -4,13 +4,22 @@ title: Default adapter usage
 until: '4.0.0'
 since: '3.12.0'
 ---
-## deprecate store.defaultAdapter (-json-api) and the -json-api adapter fallback behavior
-A deprecation warning will be shown when the store's `defaultAdapter` is accessed.
+## Deprecates both store.defaultAdapter (which defaults to -json-api) and the -json-api adapter fallback behavior
 
-To remove this deprecation warning, create an `app/adapters/application.js` with the following:
+Previously, applications could define the store.adapter property which would be used by defaultAdapter and adapterFor as a fallback for when an adapter was not found by an exact name match.
 
-    import { JSONAPIAdapter } from '@ember-data/adapters';
+[RFC 522](https://github.com/emberjs/rfcs/pull/522) Deprecated specifying and using these properties in favor of an explicit application adapter fallback.
 
-    export default JSONAPIAdapter.extend();
+If you were not setting this value previously, the following should be sufficient to resolve this deprecation:
 
-More information about custom adapters can be found on the [ember.js/guides](https://guides.emberjs.com/release/models/customizing-adapters/)
+create the file app/adapters/application.js with the following:
+
+export { default } from '@ember-data/adapters/json-api';
+
+
+If you were setting the adapter property previously to <adapter-name>, create the file app/adapters/application.js with the following:
+
+export { default } from './<adapter-name>';
+
+
+ More information about custom adapters can be found on the [ember.js/guides](https://guides.emberjs.com/release/models/customizing-adapters/) and in the [API DOCs](https://api.emberjs.com/ember-data/release/modules/@ember-data%2Fadapter)
--- a/content/ember-data/v3/default-adapter.md
+++ b/content/ember-data/v3/default-adapter.md
@@ -7,4 +7,8 @@ since: '3.12.0'
 ## deprecate store.defaultAdapter (-json-api) and the -json-api adapter fallback behavior
 A deprecation warning will be shown when the store's `defaultAdapter` is accessed.
 
-To remove this deprecation warnings, users should implement an application adapter.
+To remove this deprecation warnings, create an `app/adapters/application.js` with the following:
+
+    import { JSONAPIAdapter } from '@ember-data/adapters';
+
+    export default JSONAPIAdapter.extend();
--- a/content/ember-data/v3/default-adapter.md
+++ b/content/ember-data/v3/default-adapter.md
@@ -8,7 +8,7 @@ since: '3.12.0'
 
 Previously, applications could define the store.adapter property which would be used by defaultAdapter and adapterFor as a fallback for when an adapter was not found by an exact name match.
 
-[RFC 522](https://github.com/emberjs/rfcs/pull/522) Deprecated specifying and using these properties in favor of an explicit application adapter fallback.
+[RFC 522](https://github.com/emberjs/rfcs/pull/522) Deprecated specifying and using this property in favor of an explicit application adapter fallback.
 
 If you were not setting this value previously, the following should be sufficient to resolve this deprecation:
 
--- a/content/ember-data/v3/default-serializer.md
+++ b/content/ember-data/v3/default-serializer.md
@@ -9,7 +9,7 @@ since: '3.12.0'
 Previously, if no application or type-specific serializer was specified, the store would attempt to lookup a serializer via the `defaultSerializer` property on the type's adapter. This behavior is deprecated in favor of explicitly defining a type-specific serializer or application serializer as described in "clearing these deprecations" below.
 
 ### -default serializer fallback in store.serializerFor
-When absolutely no serializer has been defined by the app, i.e. there is no model, application, or adapter serializer, the default (`-json-api`) serializer must be used. This behavior is deprecated in favor of explicitly defining a serializer for the type or application as a whole as described below.
+Previously, when no type-specific serializer, application serializer, or adapter defaultSerializer had been defined by the app, the `-default` serializer would be used which defaulted to `-json-api`. This behavior is deprecated in favor of explicitly defining an application or type-specific serializer as described below.
 
 ## clearing these deprecations
 More information about custom serializers can be found on the [ember.js/guides](https://guides.emberjs.com/release/models/customizing-serializers/#toc_customizing-serializers)
--- a/content/ember-data/v3/default-serializer.md
+++ b/content/ember-data/v3/default-serializer.md
@@ -4,12 +4,28 @@ title: Default serializers usage
 until: '4.0.0'
 since: '3.12.0'
 ---
-## deprecate adapter.serializer and adapter.defaultSerializer fallbacks
-If no application and type specific serializer is specified, the default serializer from the type's adapter is used. This behavior is now deprecated.
+## Default serializer derprecations
+### deprecate adapter.serializer and adapter.defaultSerializer fallbacks
+If no application and type specific serializer is specified, the store will lookup the default serializer from the type's adapter. This behavior is now deprecated. This behavior is deprecated in favor of explicitly defining a serializer for the type or application as a whole as described below.
 
-To clear this deprecation users should implement an application or type specific serializer.
+### -default serializer fallback in store.serializerFor
+When absolutely no serializer has been defined by the app, i.e. there is no model, application, or adapter serializer, the default (`-json-api`) serializer must be used. This behavior is deprecated in favor of explicitly defining a serializer for the type or application as a whole as described below.
 
-## -default serializer fallback in store.serializerFor
-When no model, application or adapter serializer has specified and the default must be used.
+## clearning the deprecations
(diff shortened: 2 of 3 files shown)


content/ember-data/v3/default-adapter.md
```diff
@@ -0,0 +1,10 @@
+---
+id: ember-data:default-adapter
+title: Default adapter usage
+until: '4.0.0'
+since: '3.12.0'
+---
+## deprecate store.defaultAdapter (-json-api) and the -json-api adapter fallback behavior
+A deprecation warning will be shown when the store's `defaultAdapter` is accessed.
+
+To remove this deprecation warnings, users should implement an application adapter.
```
rwjblue: I think we need more details here. Specifically, we _know_ what **exact** thing to tell them to do but this message is pretty vague. We know they were using the default adapter, and we know what that adapter actually is, which means that we can tell them something like:

```suggestion
To remove this deprecation warnings, create an `app/adapters/application.js` with the following:

    import { JSONAPIAdapter } from '@ember-data/adapters';

    export default JSONAPIAdapter.extend();
```

(I am not 100% sure on the import location for this, please double check / confirm)

content/ember-data/v3/default-serializer.md
```diff
@@ -0,0 +1,15 @@
+---
+id: ember-data:default-serializers
+title: Default serializers usage
+until: '4.0.0'
+since: '3.12.0'
+---
+## deprecate adapter.serializer and adapter.defaultSerializer fallbacks
+If no application and type specific serializer is specified, the default serializer from the type's adapter is used. This behavior is now deprecated.
+
+To clear this deprecation users should implement an application or type specific serializer.
```
rwjblue: Same as above, we need more information. Before / after, what to do specifically, etc.

content/ember-data/v3/default-serializer.md
```diff
@@ -0,0 +1,15 @@
...
+
+## -default serializer fallback in store.serializerFor
+When no model, application or adapter serializer has specified and the default must be used.
+
+To clear this deprecation users should implement their an application serializer.
```
rwjblue: Same as above, I think we know **exactly** what they need to do, we should tell them...

content/ember-data/v3/toJSON.md
```diff
@@ -0,0 +1,10 @@
+---
+id: record.toJSON
+title: Record toJSON usage
+until: '4.0.0'
+since: '3.12.0'
+---
+
+Users should avoid calling [`toJSON`](https://github.com/emberjs/data/blob/1be481a4924b2b4316c1cc151a58328c88903dcd/packages/store/addon/-private/system/model/model.js#L620) on a record instance since it uses the now deprecated `-default` serializer to create a JSON representation of a model.
+
+To clear this deprecation users may call record.serialize() or implement their own toJSON instead.
```
rwjblue: Please add a before / after code snippet to make this a bit clearer.

rwjblue:
Also, cc @runspired for visibility / review.

content/ember-data/v3/default-adapter.md
pete-the-pete: Yep, that's the correct location: https://guides.emberjs.com/release/models/customizing-adapters/

pete-the-pete:
> Thanks for working on this!

👍 Thanks for reviewing and offering suggestions. i'll put some time into this now.

content/ember-data/v3/default-adapter.md
```diff
@@ -0,0 +1,16 @@
+---
+id: ember-data:default-adapter
+title: Default adapter usage
+until: '4.0.0'
+since: '3.12.0'
+---
+## deprecate store.defaultAdapter (-json-api) and the -json-api adapter fallback behavior
+A deprecation warning will be shown when the store's `defaultAdapter` is accessed.
+
+To remove this deprecation warnings, create an `app/adapters/application.js` with the following:
```
rwjblue:
```suggestion
To remove this deprecation warning, create an `app/adapters/application.js` with the following:
```

content/ember-data/v3/default-serializer.md
```diff
@@ -0,0 +1,31 @@
...
+since: '3.12.0'
+---
+## Default serializer derprecations
+### deprecate adapter.serializer and adapter.defaultSerializer fallbacks
+If no application and type specific serializer is specified, the store will lookup the default serializer from the type's adapter. This behavior is now deprecated. This behavior is deprecated in favor of explicitly defining a serializer for the type or application as a whole as described below.
+
+### -default serializer fallback in store.serializerFor
+When absolutely no serializer has been defined by the app, i.e. there is no model, application, or adapter serializer, the default (`-json-api`) serializer must be used. This behavior is deprecated in favor of explicitly defining a serializer for the type or application as a whole as described below.
+
+## clearning the deprecations
```
rwjblue:
```suggestion
## clearing the deprecations
```

content/ember-data/v3/toJSON.md
pete-the-pete: @runspired can you take a look at the code examples I added?

content/ember-data/v3/default-serializer.md
```diff
@@ -0,0 +1,31 @@
+---
+id: ember-data:default-serializers
+title: Default serializers usage
+until: '4.0.0'
+since: '3.12.0'
+---
+## Default serializer derprecations
```
runspired:
```suggestion
## Default serializer deprecations
```

content/ember-data/v3/default-serializer.md
```diff
@@ -0,0 +1,31 @@
...
+### deprecate adapter.serializer and adapter.defaultSerializer fallbacks
```
runspired: I am 100% confident that `adapter.serializer` was never a working codepath and we've now removed it. I think we can just say this is for `adapter.defaultSerializer`

content/ember-data/v3/default-serializer.md
```diff
@@ -0,0 +1,31 @@
...
+If no application and type specific serializer is specified, the store will lookup the default serializer from the type's adapter. This behavior is now deprecated. This behavior is deprecated in favor of explicitly defining a serializer for the type or application as a whole as described below.
```
runspired:
```suggestion
Previously, if no application or type-specific serializer was specified, the store would attempt to lookup a serializer via the `defaultSerializer` property on the type's adapter. This behavior is deprecated in favor of explicitly defining a type-specific serializer or application serializer as described in "clearing these deprecations" below.
```

content/ember-data/v3/default-serializer.md
```diff
@@ -0,0 +1,31 @@
...
+
+### -default serializer fallback in store.serializerFor
+When absolutely no serializer has been defined by the app, i.e. there is no model, application, or adapter serializer, the default (`-json-api`) serializer must be used. This behavior is deprecated in favor of explicitly defining a serializer for the type or application as a whole as described below.
+
+## clearing the deprecations
```
runspired:
```suggestion
## clearing these deprecations
```

content/ember-data/v3/default-serializer.md
```diff
@@ -0,0 +1,31 @@
...
+title: Default serializers usage
+until: '4.0.0'
+since: '3.12.0'
+---
+## Default serializer derprecations
+### deprecate adapter.serializer and adapter.defaultSerializer fallbacks
+If no application and type specific serializer is specified, the store will lookup the default serializer from the type's adapter. This behavior is now deprecated. This behavior is deprecated in favor of explicitly defining a serializer for the type or application as a whole as described below.
+
+### -default serializer fallback in store.serializerFor
+When absolutely no serializer has been defined by the app, i.e. there is no model, application, or adapter serializer, the default (`-json-api`) serializer must be used. This behavior is deprecated in favor of explicitly defining a serializer for the type or application as a whole as described below.
```
runspired:
```suggestion
Previously, when no type-specific serializer, application serializer, or adapter defaultSerializer had been defined by the app, the `-default` serializer would be used which defaulted to `-json-api`. This behavior is deprecated in favor of explicitly defining an application or type-specific serializer as described below.
```

content/ember-data/v3/default-serializer.md
```diff
@@ -0,0 +1,31 @@
...
+
+## clearing the deprecations
+More information about custom serializers can be found on the [ember.js/guides](https://guides.emberjs.com/release/models/customizing-serializers/#toc_customizing-serializers)
```
runspired:
```suggestion
More information about custom serializers can be found in the [Serializer API Docs](https://api.emberjs.com/ember-data/release/modules/@ember-data%2Fserializer) or on the [ember.js/guides](https://guides.emberjs.com/release/models/customizing-serializers/#toc_customizing-serializers)
```

content/ember-data/v3/default-serializer.md
```diff
@@ -0,0 +1,31 @@
...
+
+If a specific model type requires custom serialization, a type-specific serializer can be created.. Defining a type-specific serializer will clear the first deprecation, but won't clear both deprecations unless a type-specific serializer is defined for every type of model used in the application. To do this, create an `app/serializers/[type].js` with the following:
+
+    import { JSONAPISerializer } from '@ember-data/serializers';
+
+    export default JSONAPISerializer.extend({/* custom behavior */ });
```
runspired: Missing backticks and language for these code samples, and we should use class syntax e.g.

```ts
export default class UserSerializer extends JSONApiSerializer {
  /* custom behavior */
}
```

content/ember-data/v3/default-serializer.md
```diff
@@ -0,0 +1,31 @@
...
+
+Defining a serializer for the entire application will clear the second deprecation, and can clear both deprecations if the app doesn't require any type-specific serialization. To do this, create an `app/serializers/application.js` with the following:
+
+    import { JSONAPISerializer } from '@ember-data/serializers';
```
runspired: backticks and class syntax

content/ember-data/v3/toJSON.md
```diff
@@ -0,0 +1,44 @@
...
+
+Users should avoid calling [`toJSON`](https://github.com/emberjs/data/blob/1be481a4924b2b4316c1cc151a58328c88903dcd/packages/store/addon/-private/system/model/model.js#L620) on a record instance since it uses the now deprecated `-default` serializer to create a JSON representation of a model.
+
+To clear this deprecation users may call record.serialize() or implement their own toJSON instead. The simplest 1:1 refactor is to import a serializer and define a `toJSON` method that returns the serialized data from the model, but users may want to consider implementing a custom "serialize" method that outputs relevant data.
+
+An example of the simple refactor is below:
+### before
+
+  `app/models/post.js`
+
```
runspired: missing backticks

content/ember-data/v3/toJSON.md
```diff
@@ -0,0 +1,44 @@
...
+    import Model from '@ember-data/model';
+
+    export default Model.extend({});
+
+  `other app code`
+
+    const record = store.peekRecord('post');
```
runspired: missing backticks

content/ember-data/v3/toJSON.md
```diff
@@ -0,0 +1,44 @@
...
+    // users the default serializer, will have a deprecation warning
+    const output = record.toJSON();
+
+### after
+  `app/models/post.js`
+
+    import Model from '@ember-data/model';
```
runspired: missing backticks

content/ember-data/v3/toJSON.md
```diff
@@ -0,0 +1,44 @@
...
+    import { JSONAPISerializer } from '@ember-data/serializers';
+
+    export default Model.extend({
+      toJSON(options) {
+        let snapshot = this._internalModel.createSnapshot();
+        return JSONAPISerializer.serialize(snapshot, options);
```
runspired: A few things:

1) This is attempting to call serialize as a static class method instead of as an instance method. `this.store.serializerFor` vs importing and calling is the way to go.

2) If we must show how to do this with private API then we would use `this._createSnapshot()` instead of via `_internalModel`. However, instead of recommending a private API we could recommend calling `this.serialize()` and munging the response as desired or iterating the record directly via `eachAttribute` `eachRelationship` etc.

Another alternative is to recommend using `recordDataFor` although I think this may not work yet (we should make sure this works nicely for folks in the near future though)

content/ember-data/v3/default-serializer.md
pete-the-pete: Ha, "derp".

content/ember-data/v3/default-adapter.md
```diff
@@ -0,0 +1,29 @@
...
+id: ember-data:default-adapter
+title: Default adapter usage
+until: '4.0.0'
+since: '3.12.0'
+---
+## Deprecates both store.defaultAdapter (which defaults to -json-api) and the -json-api adapter fallback behavior
+
+Previously, applications could define the store.adapter property which would be used by defaultAdapter and adapterFor as a fallback for when an adapter was not found by an exact name match.
+
+[RFC 522](https://github.com/emberjs/rfcs/pull/522) Deprecated specifying and using these properties in favor of an explicit application adapter fallback.
```
runspired:
```suggestion
[RFC 522](https://github.com/emberjs/rfcs/pull/522) Deprecated specifying and using this property in favor of an explicit application adapter fallback.
```

content/ember-data/v3/default-serializer.md
```diff
@@ -0,0 +1,33 @@
+---
+id: ember-data:default-serializers
+title: Default serializers usage
+until: '4.0.0'
+since: '3.12.0'
+---
+## Default serializer deprecations
+### deprecate adapter.defaultSerializer fallbacks
```
runspired:
```suggestion
### deprecate adapter.defaultSerializer fallback
```

content/ember-data/v3/default-serializer.md
```diff
@@ -0,0 +1,33 @@
...
+
+### -default serializer fallback in store.serializerFor
+Previously, when no type-specific serializer, application serializer, or adapter defaultSerializer had been defined by the app, the `-default` serializer would be used which defaulted to `-json-api`. This behavior is deprecated in favor of explicitly defining an application or type-specific serializer as described below.
+
+## clearing these deprecations
+More information about custom serializers can be found in the [Serializer API Docs](https://api.emberjs.com/ember-data/release/modules/@ember-data%2Fserializer) or on the [ember.js/guides](https://guides.emberjs.com/release/models/customizing-serializers/#toc_customizing-serializers)
+
+If a specific model type requires custom serialization, a type-specific serializer can be created.. Defining a type-specific serializer will clear the first deprecation, but won't clear both deprecations unless a type-specific serializer is defined for every type of model used in the application. To do this, create an `app/serializers/[type].js` with the following:
```
runspired:
```suggestion
If a specific model type requires custom serialization, a type-specific serializer can be created. A single `application` serializer can be used a for any model types not requiring custom serialization. To define a type-specific serializer, create an `app/serializers/[type].js` with the following:
```

content/ember-data/v3/default-serializer.md
```diff
@@ -0,0 +1,33 @@
...
+
+```js
+    import { JSONAPISerializer } from '@ember-data/serializers';
+
+    export default class UserSerializer extends JSONApiSerializer { /*custom code*/ };
```
runspired:
```suggestion
    export default class UserSerializer extends JSONApiSerializer {
      /*custom code*/
    };
```

runspired:
```suggestion
    export default class UserSerializer extends JSONApiSerializer {
      /*custom code*/
    }
```

content/ember-data/v3/default-serializer.md
```diff
@@ -0,0 +1,33 @@
...
+```
+
+Defining a serializer for the entire application will clear the second deprecation, and can clear both deprecations if the app doesn't require any type-specific serialization. To do this, create an `app/serializers/application.js` with the following:
```
runspired:
```suggestion
Defining a serializer for the entire application can be done by adding the file `app/serializers/application.js` with the following:
```

content/ember-data/v3/toJSON.md
```diff
@@ -0,0 +1,46 @@
+---
+id: record.toJSON
+title: Record toJSON usage
+until: '4.0.0'
+since: '3.12.0'
+---
+
+Users should avoid calling [`toJSON`](https://github.com/emberjs/data/blob/1be481a4924b2b4316c1cc151a58328c88903dcd/packages/store/addon/-private/system/model/model.js#L620) on a record instance since it uses the now deprecated `-default` serializer to create a JSON representation of a model.
```
runspired: Do we need a header here?

runspired:
```suggestion
Previously users could use  [`record.toJSON`](https://github.com/emberjs/data/blob/1be481a4924b2b4316c1cc151a58328c88903dcd/packages/store/addon/-private/system/model/model.js#L620) to get a simple JSON serialization of a record instance by either calling the method directly or using `JSON.stringify(record)`.

This method used the now deprecated `-json` serializer to create this JSON representation of the record instead of the user supplied serializer. In addition to the surprising use of a different serializer, this creates an unnecessary dependency on the `JSONSerializer` for applications that may not otherwise have imported and uses this serializer.

We have deprecated EmberData's own implementation of this method in favor of users implementing their own (or refactoring away).
```

content/ember-data/v3/default-serializer.md
runspired: I felt the way this read before might accidentally lead some users to think they needed to implement a serializer for every single model, so I reworded it to make it clear only those types needing customization needed their own serializers and that otherwise using the application serializer was enough

content/ember-data/v3/default-adapter.md
```diff
@@ -0,0 +1,29 @@
...
+If you were not setting this value previously, the following should be sufficient to resolve this deprecation:
+
+create the file app/adapters/application.js with the following:
+
+```js
+    export { default } from '@ember-data/adapters/json-api';
+```
+
+
+If you were setting the adapter property previously to <adapter-name>, create the file app/adapters/application.js with the following:
```
rwjblue: I think we need to wrap the `<adapter-name>` thing in backticks because otherwise some markdown engines will strip it ("scary userland html").

```suggestion
If you were setting the adapter property previously to `<adapter-name>`, create the file app/adapters/application.js with the following:
```

PR #458, opened 2019-11-03
Website redesign rfc

jenweber:
The goal of this PR is to minimally apply the website redesign styles found in ember-styleguide (https://github.com/ember-learn/ember-styleguide/pull/145)

It aims to remove more CSS than it adds, and serve as a blank canvas to apply more advanced styles to after initial launch.

## Changed
- colors to match the styleguide
- We skipped h3 in our templates. I fixed that, and then I made it so that all articles start at h4 or lower. Also fixed some weirdly rendered ember data things where a title followed `====`
- use grid system for sidebar layout
- simplified the side nav

## Removed
- almost all the CSS. It was mostly cruft from the old website styles, pre-ember-styleguide.

## Needs work to reach MVP
- Does not yet have mobile view
- default link color is not accessible for regular sized text
- The search dropdowns need a lot of work, probably.
- I lost the little language arrows, and I think they are stacked at the bottom of the page somewhere out in space, forcing a big empty area at the bottom of the page. Such mystery.
- versions on the home page should be converted to cards
- fix the footer

## Temporary changes

I pinned to a specific SHA of ember-styleguide. This would need too be replaced with the correct published version.

## Screenshots

Screen Shot 2019-11-03 at 12 00 48 AM

Screenshot_2019-11-03 Ember js - Deprecations


jaredgalanis:
I’ll be trying to advance this one this weekend.

jaredgalanis:
To Dos: 
- [ ] versions on the home page should be converted to cards
- [ ] little language arrows
- [ ] mobile toc button not showing on safari
- [ ] animations for toc
- [ ] loading spinner / animations for route transitions?

jenweber:
Some a11y considerations:
- the mobile quick jump nav is very low in the DOM so you can only tab to it after tabbing through everything else. I think it should probably come in the same tab order as the left nav does - right after search. The easiest way to achieve this is to actually put it in the DOM at that position. Ember-elsewhere is a convenient hack if something about the app's data structure makes the menu tough to relocate.
- The button needs an `aria-expanded` attribute, I think. See http://web-accessibility.carnegiemuseums.org/code/navigation/
- The element wrapping the mobile nav UL should probably be `<nav>`. It was a div in the existing deployed app but hopefully correcting this will have no impact on styling. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav

The styleguide navbar itself has a lot of thought put into a11y and it's a good example of the kinds of things the mobile nav will need as well.

I like the design of this and I think it's a massive improvement over the current mobile experience, which looks broken.

jaredgalanis:
> Some a11y considerations:
> 
> * the mobile quick jump nav is very low in the DOM so you can only tab to it after tabbing through everything else. I think it should probably come in the same tab order as the left nav does - right after search. The easiest way to achieve this is to actually put it in the DOM at that position. Ember-elsewhere is a convenient hack if something about the app's data structure makes the menu tough to relocate.
> * The button needs an `aria-expanded` attribute, I think. See http://web-accessibility.carnegiemuseums.org/code/navigation/
> * The element wrapping the mobile nav UL should probably be `<nav>`. It was a div in the existing deployed app but hopefully correcting this will have no impact on styling. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav
> 
> The styleguide navbar itself has a lot of thought put into a11y and it's a good example of the kinds of things the mobile nav will need as well.
> 
> I like the design of this and I think it's a massive improvement over the current mobile experience, which looks broken.

Thanks for the feedback @jenweber. I made a few changes which hopefully address these concerns.

app/styles/app.scss
```diff
@@ -119,6 +115,7 @@ table {
     width: 100vw;
     background-color: #ebeef2;
     overflow-y: scroll;
+    z-index: 999;
```
MelSumner, line 128: please try to find a lower z-index that is appropriate

app/styles/app.scss
```diff
@@ -139,7 +136,7 @@ table {
     bottom: 10rem;
     right: 1rem;
     padding-top: 1rem;
-    z-index: 101;
+    z-index: 1000;
```
MelSumner, line 149: please try to find a lower z-index that is appropriate

app/index.html
```diff
@@ -5,7 +5,7 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <title>Ember.js - Deprecations</title>
     <meta name="description" content="">
-    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1.0, user-scalable=0"">
```
MelSumner, line 8:
```suggestion
    <meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=yes"">
```
`user-scalable` has to be `yes` for accessibility reasons.

app/index.html
```diff
@@ -5,7 +5,7 @@
...
     <link href="https://fonts.googleapis.com/css?family=Source+Sans+Pro:300,400,600" rel="stylesheet">
     <link rel="stylesheet" href="https://glimmer-styleguide.global.ssl.fastly.net/glimmer-styleguide/master/app-cf72b59a166beb307683d9d0dae6bb0f.css" />
```
MelSumner, line 10: We shouldn't need this stylesheet anymore

app/index.html
```diff
@@ -5,7 +5,7 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <title>Ember.js - Deprecations</title>
     <meta name="description" content="">
-    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1.0, user-scalable=0"">
     <link href="https://fonts.googleapis.com/css?family=Source+Sans+Pro:300,400,600" rel="stylesheet">
```
MelSumner, line 9: We shouldn't need this stylesheet anymore, it should all come from `ember-styleguide`

app/templates/application.hbs
```diff
@@ -1,18 +1,15 @@
...
+
+<main>
+  <div class="container">
+    {{outlet}}
+  </div>
+</main>
+
+<EsFooter />
 
 <script src="https://cdnjs.cloudflare.com/ajax/libs/custom-elements/1.0.0/custom-elements.min.js"></script>
```
MelSumner, line 15: Do we know what this is here for?

public/images/search-icon.svg
```diff
@@ -1,7 +1,11 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<?xml version="1.0" encoding="UTF-8"?>
 <svg width="16px" height="16px" viewBox="0 0 16 16" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
-    <defs></defs>
+    <!-- Generator: Sketch 62 (91390) - https://sketch.com -->
+    <title>Group</title>
```
MelSumner, line 4: This should be a meaningful title

public/images/search-icon.svg
```diff
@@ -1,7 +1,11 @@
...
+    <desc>Created with Sketch.</desc>
```
MelSumner, line 5: This should be a meaningful description

app/templates/application.hbs
jaredgalanis: It appears that it's a script that helps avoid Cloudflare apps from namespace collisions with the website they are installed on. 
https://www.cloudflare.com/apps/developer/docs/techniques/styles
I'm unaware of a Cloudflare app being installed on the Deprecations app so we probably don't need it. Removing it doesn't appear to break anything so I'll remove it for now.

Thanks for finding this!

app/templates/index.hbs
```diff
@@ -1,48 +1,50 @@
...
+      You can use newer versions of the CLI with older Ember apps. It's a good
+      idea to resolve deprecations as they arise so that when a major version is
+      available, it's easier to upgrade your app. For some significant changes,
+      codemods may be available, so check out
+      <a href="https://emberjs.com/blog/">
+        Release Blog Posts
+      </a>
+      before you dive in.
+    </p>
+    <h3 class="anchorable-toc" id="toc_ember">Versions</h3>
```
MelSumner, line 25:
```suggestion
    <h2 class="anchorable-toc" id="toc_ember">Versions</h2>
```

app/styles/_ember-version-graphic.scss
```diff
@@ -3,9 +3,9 @@
   $white-color: #FFFFFF;
 
   $mascots: (
-    tomster-deprecated: 'https://emberjs.com/images/tomsters/deprecated-24565384.png',
-    tomster-lts: 'https://emberjs.com/images/tomsters/lts-8974c71c.png',
-    zoey: 'https://emberjs.com/images/tomsters/zoey-2f531686.png'
+    tomster-deprecated: '',
```
MelSumner, line 6: Where are these?

app/templates/application.hbs
```diff
@@ -1,18 +1,14 @@
-{{set-body-class "deprecations deprecations_v2 responsive"}}
+<EsHeader>
+  <form class="search-form">
+    <SearchInput />
```
MelSumner, line 3: We need to put a label in here for the search input, even if we visually hide the label.

app/styles/_ember-version-graphic.scss
jaredgalanis: yeah this was just a quick hack to remove them because the links to those images are broken and it was showing up as an error in the browser console. i'm not sure if they were intentionally broken, but I think if they worked they'd show some version of the tomsters in the version cards. not sure how that fits with the new design.

app/templates/components/table-of-contents.hbs
```diff
@@ -1,24 +1,29 @@
...
-              </a>
-            </li>
-          {{/each}}
-        </ol>
-      {{/panel.body}}
-    {{/panels.panel}}
-  {{/each}}
-{{/cp-panels}}
\ No newline at end of file
+<div class="padding-vertical-small bg-light-muted rounded-sm">
```
MelSumner, line 1: I'm not super sure what's going on when this component is rendered, but it's wrapping each `<li>` element in an `<a>` which is giving us a nested interactive element violation.

app/styles/_ember-version-graphic.scss
MelSumner: It's probably because of the fingerprinting. I'd try it without that and see if it works?

app/templates/components/table-of-contents.hbs
```diff
@@ -1,24 +1,29 @@
...
+  {{#cp-panels accordion=true as |panels|}}
+    <div class="large padding-bottom-small">
+      Jump to:
```
MelSumner, line 4:
```suggestion
      Deprecations by Version:
```

app/styles/_ember-version-graphic.scss
jaredgalanis: 💯

jaredgalanis:
@MelSumner I think I've addressed your comments now. Thanks for the review!

jaredgalanis:
> Looking at the deprecations side bar-
> image
> 
> We can see that there are nested interactive elements.
> How about-
> A toggle (button) that is the sibling of the dropdown menu. That way we don't run into the accessibility issues.

Gotcha, yeah that makes sense. I _tried_ to accomplish your suggestion the latest few commits.

Issue #533, opened 2020-03-02
Incorrect example in router-events.md


JoshCrozier:
While resolving some deprecation warnings, I noticed that the example in [router-events.md](https://github.com/ember-learn/deprecation-app/blob/master/content/ember/v3/router-events.md) is incorrect.

Currently the event listeners are _not_ attached to `this.router`:

```js
this.on('routeWillChange', transition => {
  // ..
});

this.on('routeDidChange', transition => {
  // ..
});
```

Should be:
```js
this.router.on('routeWillChange', transition => {
  // ..
});

this.router.on('routeDidChange', transition => {
  // ..
});
```

I have a PR incoming for this quick documentation fix.

jaredgalanis:
@JoshCrozier thanks for looking into this. I'll be happy to review your PR when it's available.

PR #534, opened 2020-03-02
Fix the example in router-events.md

JoshCrozier:
This PR resolves issue #533 - it addresses the incorrect documentation example in [router-events.md](https://github.com/ember-learn/deprecation-app/blob/master/content/ember/v3/router-events.md).

### DIFF
--- a/content/ember/v3/router-events.md
+++ b/content/ember/v3/router-events.md
@@ -45,18 +45,19 @@ import Router from '@ember/routing/router';
 import { inject as service } from '@ember/service';
 
 export default Router.extend({
+  router: service(),
   currentUser: service('current-user'),
 
   init() {
     this._super(...arguments);
-    this.on('routeWillChange', transition => {
+    this.router.on('routeWillChange', transition => {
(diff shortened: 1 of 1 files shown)


content/ember/v3/router-events.md
```diff
@@ -45,18 +45,19 @@ import Router from '@ember/routing/router';
 import { inject as service } from '@ember/service';
 
 export default Router.extend({
+  router: service(),
```
jaredgalanis: @JoshCrozier thank you for the PR! 

I wonder if it's appropriate to inject the Router service into an extension of the Router here? 

I could see how we might want to inject the Router service if we were on a Route as is done [here](https://github.com/emberjs/ember.js/pull/17553/files), but that doesn't seem to be the case in these examples.

What do you think?

JoshCrozier:
@jaredgalanis - Ah yes, you are completely correct.

When I was reading though the documentation I was under the impression they were all examples for the `Route` because of [the example](https://deprecations.emberjs.com/v3.x/#toc_remove-handler-infos) directly above this one.

Issue #535, opened 2020-03-04
Make explanation of Component#isVisible deprecation more explicit


locks:
The deprecation prose for [Component#isVisible](https://deprecations.emberjs.com/v3.x#toc_ember-component-is-visible) is somewhat terse and should be improved with a more detailed migration.

locks:
Assigned to @vinodhrajamuralidharan.

jessica-jordan:
Closing via #536

PR #536, opened 2020-03-04
isVisible content update

vinodhrajamuralidharan:
isVisible content update

https://github.com/ember-learn/deprecation-app/issues/535

### DIFF
--- a/content/ember/v3/component-is-visible.md
+++ b/content/ember/v3/component-is-visible.md
@@ -1,6 +1,10 @@
+---
+id: ember-component.is-visible
+title: Deprecate `Component#isVisible`
 until: '4.0.0'
 since: '3.15'
 ---
+
 In ember earlier versions, we had the concept of 'A controller should never call method or change property on its associated view '( Ember.View class)', instead the view should bind the state of its associated controller.'
 
 So, we were using something like this,
--- a/content/ember/v3/component-is-visible.md
+++ b/content/ember/v3/component-is-visible.md
@@ -5,6 +5,7 @@ In ember earlier versions, we had the concept of 'A controller should never call
 
 So, we were using something like this,
 ```js
+import Ember from 'ember';
 App.MyView = Ember.View.extend({
   isVisible: Ember.computed.alias('controller.myViewVisible')
 }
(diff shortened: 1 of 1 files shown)


content/ember/v3/component-is-visible.md
```diff
@@ -1,21 +1,48 @@
...
+      return A(['john', 'peter']);
+    }
+```
+{{! app/templates/components/my-question.hbs }}
+{{! `if` helper }}
+{{#if this.students.length}}
   <MyComponent />
 {{/if}}
+```
+or in the html element, we can use [`hidden`] (https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) attribute and pass the value dynamically:
```
jessica-jordan: I think it'd be
```suggestion
or in the html element, we can use the [`hidden`] (https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) attribute and pass the value dynamically:
```

content/ember/v3/component-is-visible.md
```diff
@@ -1,21 +1,48 @@
...
+  {{! app/templates/components/my-question.js }}
+  get hideMe() {
+    return true;
+  }
 ```
 
+```hbs
+  {{! app/templates/components/my-question.hbs }}
+  {{! `hidden` attribute }}
+  <div hidden={{this.hideMe}}></div>
```
jessica-jordan: Nice example, by the way :sparkles:

content/ember/v3/component-is-visible.md
```diff
@@ -1,21 +1,48 @@
...
+So, we were using something like this,
+```js
+App.MyView = Ember.View.extend({
+  isVisible: Ember.computed.alias('controller.myViewVisible')
+}
+```
+
+or in the component
+
+```js
```
jessica-jordan: Would it make sense to add the context of the component into this example as well? f.e. by adding the import and export statements for the component class - what do you think?

vinodhrajamuralidharan: can you check now?

jessica-jordan: This sounds great!

vinodhrajamuralidharan:
@jessica-jordan assign me some tasks like code, documentation, infrastructure, maintenance, triaging. I will do whenever I have free time.

PR #548, opened 2020-03-22
Rewrite EmberComponent#isVisible guide with three alternatives

locks:
I rewrote the guide to use a common base component for all examples, and added a Glimmer component suggestion given this was deprecated in 3.15 which is Octane by default.
I tried coming up with a self-contained use case, hence the component hiding itself. I am not sure how `isVisible` is actually used in practice, so please comment if you are looking for guidance on updating your usage.

In the `attributeBindings` example I flipped the logic so I don't have to explain the microsyntax of `attributeBindings: ['shouldShow::hide']`, and I used `attributeBindings` so it is applicable to any tagful component.
I did not mention tagless classic components because that's already a more advanced usage and didn't want to muddle the explanation.

Lastly, our general recommendation is that we use classic syntax for classic components, hence my use of `{{action}}` in the example, as that is likely what the reader is also using.

### DIFF
--- a/content/ember/v3/component-is-visible.md
+++ b/content/ember/v3/component-is-visible.md
@@ -11,6 +11,8 @@ One of them is `isVisible`, which controls if the component is hidden to the end
 
 You can update your component in one of two ways, you can wrap your component's template in an `{{if}}`, or you can use the [`hidden`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) HTML attribute.
 
+It's worth noting that not all visibility approaches are equal- we recommend reviewing use of `aria-hidden` as well as accessible ways to visibly hide content while still making it available to assistive technology. 
+
 Because classic components have a wrapper `div` element by default,
 it might be necessary that you do additional changes to your component so that no content is accidentally shown.
 
--- a/content/ember/v3/component-is-visible.md
+++ b/content/ember/v3/component-is-visible.md
@@ -5,60 +5,124 @@ until: '4.0.0'
 since: '3.15'
 ---
 
-In ember earlier versions, we had the concept of 'A controller should never call method or change property on its associated view '( Ember.View class)', instead the view should bind the state of its associated controller.'
+Classic components have a number of APIs to handle the wrapper `div` that they create by default.
+One of them is `isVisible`, which controls if the component is hidden to the end user or not.
+`isVisible` is now deprecated in accordance with [RFC #324](https://github.com/emberjs/rfcs/blob/master/text/0324-deprecate-component-isvisible.md).
 
-So, we were using something like this,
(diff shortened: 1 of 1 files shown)


content/ember/v3/component-is-visible.md
```diff
@@ -5,60 +5,124 @@ until: '4.0.0'
 since: '3.15'
 ---
 
-In ember earlier versions, we had the concept of 'A controller should never call method or change property on its associated view '( Ember.View class)', instead the view should bind the state of its associated controller.'
+Classic components have a number of APIs to handle the wrapper `div` that they create by default.
+One of them is `isVisible`, which controls if the component is hidden to the end user or not.
+`isVisible` is now deprecated in accordance with [RFC #324](https://github.com/emberjs/rfcs/blob/master/text/0324-deprecate-component-isvisible.md).
 
-So, we were using something like this,
+You can update your component in one of two ways, you can wrap your component's template in an `{{if}}`, or you can use the [`hidden`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) HTML attribute.
```
Gaurav0: There is also css display and visibility styles, not to mention rendering off screen for sighted users to target screen readers only. aria-hidden should also be mentioned for accessibility in the discussion. Not sure if all this is helpful or needlessly complexifies this discussion.

locks: Those are good points. I tried to stay as close as possible to the original mechanically, which apparently does `style="display: none"` which is iffy to me xD Since it's a deprecation guide, I think `aria-hidden` should be mentioned if the original and the migrated solution have different behaviors accessibility-wise?

MelSumner:
```suggestion
You can update your component in one of two ways, you can wrap your component's template in an `{{if}}`, or you can use the [`hidden`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) HTML attribute.

It's worth noting that not all visibility approaches are equal- we recommend reviewing use of `aria-hidden` as well as accessible ways to visibly hide content while still making it available to assistive technology. 
```

locks: Is there some `aria-hidden` documentation we can link to? https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-hidden_attribute says it's draft status.

MelSumner: https://www.w3.org/WAI/PF/aria/states_and_properties#aria-hidden is what it looks like,  but that link does say editor's draft. I'd have to look for the one that isn't the draft but for all practical purposes the content there is legit.

jaredgalanis: @locks I think your updates look good. I'm inclined to say we should merge it and come back to expand if necessary. What do you think?

locks:
@jaredgalanis thanks for your review! let's get it shipped :)

PR #590, opened 2020-05-17
Update to Classes

jaredgalanis:


app/app.js
```diff
@@ -1,14 +1,12 @@
-import Application from '@ember/application';
-import Resolver from './resolver';
-import loadInitializers from 'ember-load-initializers';
-import config from './config/environment';
+import Application from '@ember/application'
```
locks, line 1: semicolons were removed in this file

app/adapters/application.js
```diff
@@ -25,7 +24,7 @@ export default DS.JSONAPIAdapter.extend({
     } else if(requestType === 'query') {
       url = [modelName, `${query.path}-${query.version}.json`];
     } else {
-      return this._super(...arguments);
```
locks, line 28: Doesn't this change the behavior?

jaredgalanis: sure would. thanks for catching that :)

app/app.js
jaredgalanis: copy pasta

jaredgalanis:
> I tried doing a review commit by commit but there seem to be multiple things changing in each one so I was getting confused. If you could split non-`class` changes into separate PRs I can review and get them in more quickly!

@locks  I backed out changes that were not directly related to the class conversion where possible. I'll open another PR that moves the processing of results which I think will get rid of some visual shifting that's happening while results are processing / async loading is occurring.

Note, I have not figured out how to deal with converting the `search-input` component. There's a strange thing with dynamically importing algolia in that file if I convert it to a class. `ember-auto-import` cannot find it on the registry. I can try to work more on this later today or tomorrow, but I'm also okay with getting this work merged and then finding a solution for that issue.

I suppose we may also want to merge your #593 first and then rebase this work on top of that.

PR #593, opened 2020-05-22
prepare app for optional  anchor id

locks:


app/helpers/id-for-deprecation.js
```diff
@@ -1,7 +1,11 @@
...
-export function idForDeprecation(params) {
-  let idString = params[0];
+export function idForDeprecation([id, anchor]) {
+  if (anchor) {
+    return anchor;
+  }
+
+  let idString = id;
   idString = idString.replace(/\s+/g, '');
   let replacedIdString = idString.replace(/\.|,/g, '-');
```
jaredgalanis, line 10: My recollection is a bit hazy, but I thought during discussion it may have been raised that colons could be part of some of the id's coming from Ember Data (i.e. not only in the format @richgt proposed). If that's the case I wonder if we want to include stripping them out here just for good measure. 
```suggestion
  let replacedIdString = idString.replace(/\.|,|:/g, '-');
```

locks:
You raised a good point, we should add some tests!

locks:
You raised a good point, we should add some tests!

PR #600, opened 2020-05-28
Deprecate getWithDefault based on RFC 0554

chrisrng:
To capture https://emberjs.github.io/rfcs/0554-deprecate-getwithdefault.html
and https://github.com/emberjs/ember.js/pull/18993

### DIFF
--- a/content/ember/v3/deprecate-get-with-default.md
+++ b/content/ember/v3/deprecate-get-with-default.md
@@ -2,7 +2,7 @@
 id: ember-metal.get-with-default
 title: Use Ember getter and explicitly check for undefined
 until: '4.0.0'
-since: '3.19'
+since: '3.20'
 ---
 
 Deprecate support for `getWithDefault` in Ember's Object module (@ember/object) – both the [function](https://api.emberjs.com/ember/release/functions/@ember%2Fobject/getWithDefault) and the [class method](https://api.emberjs.com/ember/release/classes/EmberObject/methods/getWithDefault?anchor=getWithDefault) – because its expected behaviour is confusing to Ember developers.
--- a/content/ember/v3/deprecate-get-with-default.md
+++ b/content/ember/v3/deprecate-get-with-default.md
@@ -72,3 +72,7 @@ let falseValue = get(obj, 'falseValue') ?? defaultValue;
 ```
 
 This can be an option if we are aware that either `null` or `undefined` should return the default value.
+
+Please review the deprecation RFC over at
+[emberjs/rfcs](https://emberjs.github.io/rfcs/0554-deprecate-getwithdefault.html)
+for more details.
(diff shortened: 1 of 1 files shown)


content/ember/v3/deprecate-get-with-default.md
```diff
@@ -0,0 +1,78 @@
+---
+id: ember-metal.get-with-default
+title: Use Ember getter and explicitly check for undefined
+until: '4.0.0'
+since: '3.19'
+---
+
+Deprecate support for `getWithDefault` in Ember's Object module (@ember/object) – both the [function](https://api.emberjs.com/ember/release/functions/@ember%2Fobject/getWithDefault) and the [class method](https://api.emberjs.com/ember/release/classes/EmberObject/methods/getWithDefault?anchor=getWithDefault) – because its expected behaviour is confusing to Ember developers.
+
+- The API will only return the default value when the value of the property retrieved is `undefined`. This behaviour is often overlooked when using the function where a developer might expect that `null` or other _falsy_ values will also return the default value.
```
rwjblue, line 10:
```suggestion
- The API will only return the default value when the value of the property retrieved is `undefined`. This behaviour is often overlooked when using the function where a developer might expect that `null` or other _falsey_ values will also return the default value.
```

I don't really know how to spell this non-word, but we use this spelling just below so we ought to at least stay consistent.

content/ember/v3/deprecate-get-with-default.md
```diff
@@ -0,0 +1,78 @@
...
+
+let result = get(obj, 'some.key');
+if (result === undefined) {
+  result = defaultValue;
+}
+```
+
+#### Using Nullish Coalescing Operator
+
+We cannot codemod directly into the nullish coalescing operator since the expected behaviour of `getWithDefault` is to only return the default value if it is strictly `undefined`. The nullish coalescing operator accepts either `null` or `undefined` to show the default value.
```
rwjblue, line 34: Maybe add a link to nullish coalescing on MDN or something?

content/ember/v3/deprecate-get-with-default.md
```diff
@@ -0,0 +1,78 @@
...
+let undefinedValue = get(obj, 'undefinedKey') ?? defaultValue;
+
+// Returns defaultValue 1, nullValue = 1
+let nullValue = get(obj, 'nullValue') ?? defaultValue;
+
+// Returns obj's falseValue, falseValue = false
+let falseValue = get(obj, 'falseValue') ?? defaultValue;
+```
+
+This can be an option if we are aware that either `null` or `undefined` should return the default value.
```
rwjblue, line 74: Maybe a small tweak here:

```suggestion
For any given usage of `getWithDefault`, using nullish coalescing might work very well, but keep in mind that either `null` or `undefined` will return the default value.
```

content/ember/v3/deprecate-get-with-default.md
chrisrng: :)

content/ember/v3/deprecate-get-with-default.md
chrisrng: Added!

PR #619, opened 2020-06-23
Updated Percy to v2.1.3

ijlee2:
## Description

I noticed in `package.json` and the application test we currently use `v1.x` of Percy. Dependabot wouldn't have caught this because the package name changed from `ember-percy` to `@percy/ember`.

I updated the test to rely on modern testing API (let the `settled` test helper do its job and avoid updating time ourselves).

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,11 +16,12 @@
     "lint:hbs": "ember-template-lint .",
     "lint:js": "eslint .",
     "start": "ember serve",
-    "test": "ember test"
+    "test": "percy exec -- ember test"
   },
   "devDependencies": {
     "@ember/optional-features": "^1.3.0",
     "@ember/render-modifiers": "^1.0.2",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -16,12 +16,12 @@
     "lint:hbs": "ember-template-lint .",
     "lint:js": "eslint .",
     "start": "ember serve",
-    "test": "ember test"
+    "test": "percy exec -- ember test"
   },
   "devDependencies": {
     "@ember/optional-features": "^1.3.0",
     "@ember/render-modifiers": "^1.0.2",
-    "@sinonjs/fake-timers": "^6.0.1",
```
ijlee2, line 24: I think this package was used as a workaround for Percy v1.x.

jaredgalanis: 👍  yeah it can come out

PR #620, opened 2020-06-25
Update ember-cli and ember-styleguide

jaredgalanis:
- there was an issue with upgrading ember-cli to 3.18+ in apps that use the ember-styleguide
- a fix for this issue was released in ember-styleguide 4.1.0
- this bumps version of both dependencies
- adding ie 11 to the build targets was also necessary due to use of nullish coallescing in the styleguide and some external changes in support of these operators


mansona:
@jaredgalanis any idea why all the Percy snapshots are "new" snapshots? 🤔

jaredgalanis:
> @jaredgalanis any idea why all the Percy snapshots are "new" snapshots? 🤔

I don't know this for sure, but I suspect that it relates to updating Percy in a recent PR. Will continue to monitor to see if it keeps happening.

PR #669, opened 2020-09-15
Fix typo that creates confusing error when followed

cloke:
A new ember app has a deprecation warning reported by default. Following the warning takes you to this deprecation and there is a typo in the sample code. If this is followed it then introduces a new error that says the adapter does not exist.

### DIFF
--- a/content/ember-data/v3/default-adapter.md
+++ b/content/ember-data/v3/default-adapter.md
@@ -15,7 +15,7 @@ If you were not setting this value previously, the following should be sufficien
 create the file app/adapters/application.js with the following:
 
 ```js
-    export { default } from '@ember-data/adapters/json-api';
+    export { default } from '@ember-data/adapter/json-api';
 ```
 
 


ijlee2 (approved):
Thanks for the find! I'm sorry to hear that the error caused confusion when you created your new Ember app.

Which version of Ember did you install (e.g. v3.21), if I may ask?

cloke:
> Which version of Ember did you install (e.g. v3.21), if I may ask?
Yes. 3.21. 

Ember should probably be updated to include the default adapter and serializer in the blueprint as well.

ijlee2:
Cool, I've linked your comment to Discord to see if someone from the CLI team can help.

kategengler:
@cloke Can you clarify where this deprecation is triggered in a new Ember app? I just generated an app on v3.21.2 and did not see this deprecation happen. (The output of the blueprints on v3.21.2 is here https://github.com/ember-cli/ember-new-output/tree/v3.21.2)

cloke:
@kategengler I will need to start another new one, but I bet it was once I added a model / route which was literally the first thing I did after `ember new`. Now that I think about this I am not sure what the best course of action would be because it might seem weird to have a serializer / adapter file in the blueprint, but then never use ember data.

PR #693, opened 2020-10-23
Deploy build image change

jaredgalanis:


jaredgalanis:
Per @bertdeblock's suggestion 👍 🎉  blowing away `node_modules` caused Netlify to rebuild the cached build image which means we can now be on Xenial again and have deploy previews work.

PR #696, opened 2020-10-27
[RFC#236] Add guide for String prototype extensions deprecation

locks:
Part of [RFC#236 - Ember.String deprecation](https://github.com/emberjs/rfc-tracking/issues/26).
PR twin of https://github.com/emberjs/ember.js/pull/19234.

## Tasks
- [ ] deprecate String prototype extensions
  - [ ] make sure deprecation URL is published and matches


mansona (approved):

content/ember/v3/ember-string-prototype-extensions.md
```diff
@@ -0,0 +1,63 @@
...
+
+While Ember addons (`ember addon …`) have prototype extensions disabled by default, they are enabled for applications (`ember new …`) making you able to call `"Tomster".dasherize()`, for example.
+Instead of calling the method on the string, you should instead import the function from `@ember/string`.
+
+Before:
+
+```js
+let mascot = "Empress Zoey";
+
+mascot.camelize();   //=> empressZoey
```
ijlee2, line 18: For consistency:
```suggestion
mascot.camelize();   //=> "empressZoey"
```

content/ember/v3/ember-string-prototype-extensions.md
```diff
@@ -0,0 +1,63 @@
...
+  capitalize, 
+  classify, 
+  decamelize,
+  underscore, 
+  w, 
+} from "@ember/string";
+
+let mascot = "Empress Zoey";
+
+camelize(mascot);   //=> empressZoey
```
ijlee2, line 40:
```suggestion
camelize(mascot);   //=> "empressZoey"
```

content/ember/v3/ember-string-prototype-extensions.md
```diff
@@ -0,0 +1,63 @@
...
+capitalize(mascot); //=> "Empress Zoey"
+classify(mascot);   //=> "EmpressZoey"
+decamelize(mascot); //=> "empress zoey"
+underscore(mascot); //=> "empress_zoey"
+w(mascot);          //=> [ "Empress", "Zoey" ]
+```
+
+You might want to replace these methods with another library, like [lodash](https://lodash.com/).
```
ijlee2, line 48: "might want to replace" sounded a bit like the methods from `@ember/string` may be deprecated altogether. If that isn't the case, what do you think about:

```suggestion
You may also instead rely on methods from another library like [lodash](https://lodash.com/).
```

content/ember/v3/ember-string-prototype-extensions.md
```diff
@@ -0,0 +1,63 @@
...
+Keep in mind that different libraries will behave in slightly different ways, so make sure any critical `String` transformations are thoroughly tested.
+
+You can also [disable String prototype extensions](https://guides.emberjs.com/release/configuring-ember/disabling-prototype-extensions/) by editing your environment file:
+
+```js
+// config/environment.js
+ENV = {
+  EmberENV: {
+    EXTEND_PROTOTYPES: {
+      Date: false,
```
ijlee2, line 58: I saw that `Date` is present by default in a new Ember app. I think we can just show `String`, and it's implied that other keys aren't shown in the sample code.
```suggestion
```

content/ember/v3/ember-string-prototype-extensions.md
```diff
@@ -0,0 +1,63 @@
+---
+id: ember-string.prototype-extensions
+title: Deprecate String prototype extensions
+until: '4.0.0'
+since: 'Upcoming Features'
```
ijlee2, line 5: I read the deprecations RFC a bit yesterday. Do the metadata here need to follow the proposed syntax in the RFC?

ijlee2 (approved):
I thought you explained the RFC well and concisely. Thanks for working on the write-up! I left a few optional suggestions.

content/ember/v3/ember-string-prototype-extensions.md
locks: I was thinking about the same thing, but given how we currently present and use the metadata I don't think it's necessary. I have an idea for an improvement to the app, though.

content/ember/v3/ember-string-prototype-extensions.md
locks: I hedged my bet here because the entire `@ember/string` "package" is deprecated, but we split up the RFC into three deprecations to make it easier to progress.
Let's make a note to tweak it further when the last deprecation lands.

content/ember/v3/ember-string-prototype-extensions.md
locks: I personally prefer to make the code block as close to the generated output as possible, so the reader doesn't wonder why we don't show `Date: false`.

ijlee2: Yep, I think that makes sense.

locks:
Thanks for the feedback! I'm merging it in and we can continue iterating once the other deprecations are released.

PR #698, opened 2020-11-01
Refactor toc-anchor to be accessible

### LINKED ISSUE #592: toc-anchor needs to be accessible
Currently the link to a deprecation article is not accessible because it is an `<a>` wrapping an `h3`.

Example:
 
```
<div class="padding-vertical-small bg-light-muted rounded-sm">
  <a class="bg-none toc-anchor" href="#toc_ember-string-fmt">
    <h3 class="anchorable-toc" id="toc_ember-string-fmt">Ember.String.fmt</h3>
  </a>
</div>
```

A solution would separate the header and the link from each other.

neojp:
This PR attempts to close #592

- Refactored js code from `deprecation-article.js` to `deprecation-article.hbs`
- Use an icon to display the anchor while hovering the heading the same way GitHub markdown handles it

## Screenshots

Default state

image

Anchor Icon shows up while hovering the heading

image

### DIFF
--- a/app/components/deprecation-article.js
+++ b/app/components/deprecation-article.js
@@ -69,14 +69,6 @@ export default class DeprecationArticle extends Component {
       });
     }
 
-    element.querySelectorAll(".anchorable-toc").forEach(function (currentToc) {
-      let link = document.createElement('a');
-      link.classList.add('bg-none', 'toc-anchor');
-      link.setAttribute('href', `#${currentToc.getAttribute('id')}`);
-      link.appendChild(currentToc);
-      element.prepend(link);
-    });
-
     this.prism.highlight();
   }
 }
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -95,6 +95,38 @@ table {
   width: 1px;
 }
 
+// h3 headings used in deprecation-article component
+.anchorable-toc {
+  position: relative;
+
+  &:hover .toc-anchor,
+  &:target .toc-anchor {
+    opacity: 1;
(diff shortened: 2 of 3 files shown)


jenweber (dismissed):
Hey, thanks for your work on this! I noticed two things that I think can be further improved here for UX and accessibility. Ideally, the heading link feature should be available to mobile users and users of assistive tech too. Mobile users cannot hover on the link, and for screen readers, right now it is hidden with `aria-hidden`.

The [cheat sheet app](https://ember-learn.github.io/ember-octane-vs-classic-cheat-sheet/) has an example of a header link that is shown 100% of the time, and the link symbol is usable by everyone:

Screen Shot 2020-11-05 at 9 29 37 PM

What do you think of this approach?

It is great that you have moved the logic out of the JS and into the template. It's much easier to follow now!

Kennie-create:
Good job on this, I've learned about `<a>` tag and accessibility by reading over @neojp 's PR

neojp:
@jenweber sure thing. I was following the way the default GitHub markdown pages were doing it, but we can use the one from the cheat sheet app.

ijlee2:
I rebased the branch so that the pull request will be checked against GitHub Actions instead of Travis.

neojp:
@jenweber I've made the changes to match the UI of the cheat sheet app. I've removed the `aria-hidden` attribute and made it visible at all times instead of only available through hover.

image

locks (approved):
Thanks for the help!

PR #705, opened 2020-11-07
add guide for tryInvoke deprecation

locks:
Adds guide for [tryInvoke deprecation](https://emberjs.github.io/rfcs/0674-deprecate-transition-methods-of-controller-and-route.html).

I adjusted the text from the RFC to something that felt smoother, please review!

### DIFF
--- a/content/ember/v3/try-invoke.md
+++ b/content/ember/v3/try-invoke.md
@@ -1,0 +1,28 @@
+---
+id: ember-utils.try-invoke
+title: tryInvoke from @ember/utils
+until: '4.0.0'
+since: 'Upcoming Features'
+---
+
+`tryInvoke` from the `@ember/utils` package is now deprecated.
+
+In most cases, function arguments should not be optional, but in the rare occasion that an argument is optional by design, we can replace `tryInvoke` with JavaScript's [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining).
+
+Before:
+
+```javascript
+import { tryInvoke } from '@ember/utils';
+
+foo() {
+ tryInvoke(this.args, 'bar', ['baz']);
+}
+```
+
+After:
+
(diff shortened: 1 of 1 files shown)


content/ember/v3/try-invoke.md
```diff
@@ -0,0 +1,28 @@
+---
+id: ember-utils.try-invoke
+title: tryInvoke from @ember/utils
+until: '4.0.0'
+since: 'Upcoming Features'
+---
+
+`tryInvoke` from the `@ember/utils` package is now deprecated.
+
+In most cases, function arguments should not be optional, but in the rare occasion that an argument is optional by design, we can `tryInvoke` with JavaScript's [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining).
```
ijlee2, line 10:
```suggestion
In most cases, function arguments should not be optional, but in the rare occasion that an argument is optional by design, we can replace `tryInvoke` with JavaScript's [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining).
```

jsconfig.json
```diff
@@ -0,0 +1 @@
+{"compilerOptions":{"target":"es6","experimentalDecorators":true},"exclude":["node_modules","bower_components","tmp","vendor",".git","dist"]}
```
ijlee2, line 1: Can you leave a note in the pull request why this file was added?

ijlee2 (approved):
There's a word missing that needs to be added before merging the PR. Thanks for working on updating the `deprecation-app`!

jsconfig.json
locks: Ugh, it shouldn't have been. I'll configure my global git ignore!

locks:
Thank you for the swift review! Will merge once CI passes after the latest changes.

PR #706, opened 2020-11-08
Add guide for Route and Controller transition methods deprecation

locks:
The deprecation guide suggestion in the [RFC](https://emberjs.github.io/rfcs/0674-deprecate-transition-methods-of-controller-and-route.html) for `Controller#transitionToRoute` suggests an async action.
I personally avoid async actions for ember-concurrency tasks because of some bad experiences.
While the controller is not destroyed so there shouldn't be "changed after destroyed" errors, it can still have some bad interaction with hooks like `setupController` and the like.

On the other hand, async actions are likely very common in the wild and the reader of the guide might find the code familiar.
If anyone checking this issue out has thoughts, please comment here!


jaredgalanis:
If the problematic async interactions with `setupController` aren't triggered by async actions, then I'd be fine with this. That said, I'm not familiar with what those interactions are. Are you able to point me to a resource describing the issue @locks ?

ijlee2:
@locks I think async-await is a common pattern. I'd leave the code example as is, since the problems with async-await in a component is not particular to the RFC, which is more about deprecating the transition methods.

I think it may be better if you write a blog post to describe the problems with using async-await and how to solve them.

content/ember/v3/route-controller-transition-methods.md
```diff
@@ -0,0 +1,91 @@
...
+- `replaceRoute` on `Controller`
+
+Instead, the user should inject the router service in the respective class and use its methods.
+
+### Route example
+
+Before:
+
+```javascript
+  // app/route/foo.js
```
ijlee2, line 21: It seems the RFC had a typo.

```suggestion
  // app/routes/foo.js
```

jelhan:
I agree with @ijlee2. The deprecation guide should show how to migrate from `Controller.transitionTo` to `RouterService.transitionTo`. This is independent from using `async` method, promise chaining or concurrency task.

I would use the same pattern as used in the tutorial and guides to illustrate in deprecation guide unless the deprecation is specific to using `async` method, promise chaining or concurrency task. As far as I know tutorial and guides are using `async`/`await`.

In general I'm not a big fan of teaching a community solution like ember-concurrency as integral piece of the framework. And I think ember-concurrency is not part of Ember CLI blueprint for new applications. Mostly because it's another ember specific solution not well aligned with the rest of the JavaScript community.

locks:
Updated the file names to something other than `foo`. Can you review?

jelhan (approved):
Thanks a lot for starting the implementation that quickly!

ijlee2 (approved):
Looks good! Thanks for giving the controller and route examples a practical name.

ijlee2:
This Percy issue likely might have existed before. It seems that (sometimes) we don't see code blocks in Percy snapshots?

Screen Shot 2020-11-09 at 8 01 58 AM

The issue is present in other explanations of deprecations so let's address this issue separately.

Issue #716, opened 2020-11-18
Improve readability of deprecation lists?


bertdeblock:
At the moment deprecations are visually listed like so:
<img width="260" alt="Screenshot 2020-11-18 at 18 00 06" src="https://user-images.githubusercontent.com/7403183/99562266-2ff95700-29c8-11eb-8b37-28c5f4e52b95.png">

I think readability can be improved by simply re-adding the bullets:
<img width="257" alt="Screenshot 2020-11-18 at 18 00 33" src="https://user-images.githubusercontent.com/7403183/99562537-836ba500-29c8-11eb-83ac-c7a7e1070a33.png">

If you guys think this is useful, I could issue a PR if someone lets me know how it should look like visually.

ijlee2:
Hi, @bertdeblock. Thanks for suggesting the idea of adding bullets to the list. I think adding bullets does help with improving readability.

It looks like we can modify line 14:

https://github.com/ember-learn/deprecation-app/blob/07e3aba63dc322b29cf498b4c312df958bc49a80/app/templates/components/toc-section.hbs#L14

Maybe write `class="p-0"` instead, which will add bullets and (re)set padding to `0`. However, it seems we need to apply `word-break` as well?

![Screen Shot 2020-11-18 at 12 35 41 PM](https://user-images.githubusercontent.com/16869656/99572645-96678080-299a-11eb-9ba3-831a294b299b.png)

You may want to investigate if there's a built-in class (from `ember-styleguide` or an existing stylesheet in `app/styles`) that will apply `word-break: break-word;`. If not, I think we could create a CSS class in the app.

I hope this made sense. If you have questions or would like clarification, please feel free to ask!

jaredgalanis:
This also sounds great, thanks @bertdeblock! I just want to make sure we're okay with taking the time to make the changes given that the website redesign strike team may adjust the UI of this sidebar entirely in the near'ish future.

I personally still think it is worth making these changes because we don't know when the redesign efforts touch the deprecations app.

bertdeblock:
Coolio, I'll create a PR later today and you guys can still decide if you want it merged or not. Another option would be to add some space between the list items?

Example:
<img width="256" alt="Screenshot 2020-11-19 at 08 18 16" src="https://user-images.githubusercontent.com/7403183/99633906-f401d880-2a3f-11eb-96d9-48fc74fac283.png">

ijlee2:
[...]

PR #717, opened 2020-11-19
Improve readability of deprecation lists

bertdeblock:
Addresses #716.

Adds spacing between the list items of deprecation lists.
Went with this approach because it was my personal favourite (clear and clean).
Let me know if you guys prefer a different approach.

Result:
Screenshot 2020-11-19 at 17 14 45

### DIFF
--- a/app/templates/components/toc-section.hbs
+++ b/app/templates/components/toc-section.hbs
@@ -12,8 +12,8 @@
 {{#if this.open}}
   <div>
     <ul aria-labelledby={{this.id}} class="level-3 list-unstyled">
-      {{#each @result.contents as |content|}}
-        <li>
+      {{#each @result.contents as |content index|}}
+        <li class={{if index "mt-2"}}>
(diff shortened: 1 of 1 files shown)


app/templates/components/toc-section.hbs
```diff
@@ -12,8 +12,8 @@
 {{#if this.open}}
   <div>
     <ul aria-labelledby={{this.id}} class="level-3 list-unstyled">
-      {{#each @result.contents as |content|}}
-        <li>
+      {{#each @result.contents as |content index|}}
+        <li class={{if index "mt-2"}}>
```
bertdeblock, line 16: Not really proud of this approach. I would prefer something like `mb-2 last:mb-0`, but I could not find anything in `ember-styleguide`. Is there something available like this?

jaredgalanis: I don't know of anything of the top of my head, unfortunately.

jaredgalanis: I'm good with this approach though.

jaredgalanis (approved):
Thanks for doing this! 🎉

PR #718, opened 2020-11-19
Resolve all ignored template lint errors

bertdeblock:
Noticed these when doing #717. Thought I'd might as well fix these in a separate PR.

### DIFF
--- a/app/templates/components/main-layout.hbs
+++ b/app/templates/components/main-layout.hbs
@@ -9,7 +9,6 @@
 
     {{#each @sortedGroupedResults as |result|}}
       <h2 class="anchorable-toc rounded-lg padding-small">Deprecations Added in {{result.since}} </h2>
-      {{renderId}}
       {{#each result.contents as |content|}}
         <DeprecationArticle @model={{content}} @renderIdOrUntil={{@renderIdOrUntil}} />
       {{/each}}
--- a/app/templates/components/search-input.hbs
+++ b/app/templates/components/search-input.hbs
@@ -3,7 +3,6 @@
   <input
     id="search-input"
     type="search"
-    value={{value}}
     placeholder="Search deprecations"
     autocomplete="off"
     aria-label="Search"
@@ -16,7 +15,7 @@
   target="#search-input"
   targetAttachment="bottom left"
   attachment="top left"
-  constraints=_resultTetherConstraints
+  constraints=this._resultTetherConstraints
   class="ds-dropdown-results"
 }}
   {{#if this.searchService.searchResults.hits}}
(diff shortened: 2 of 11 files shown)


app/templates/components/main-layout.hbs
```diff
@@ -9,7 +9,6 @@
 
     {{#each @sortedGroupedResults as |result|}}
       <h2 class="anchorable-toc rounded-lg padding-small">Deprecations Added in {{result.since}} </h2>
-      {{renderId}}
```
bertdeblock, line 12: I'm guessing this was a leftover from testing something? Couldn't find any occurrence of `renderId` in the entire codebase.

app/templates/components/search-input.hbs
```diff
@@ -3,7 +3,6 @@
   <input
     id="search-input"
     type="search"
-    value={{value}}
```
bertdeblock, line 6: `value` was never used / set somewhere else.

app/templates/components/search-result.hbs
```diff
@@ -12,11 +12,11 @@
               {{#if index}}
                 >
               {{/if}}
-              {{{heading.value}}}
+              {{html-safe heading.value}}
```
bertdeblock, line 15: Not sure how you guys feel about adding new dependencies. Let me know if this should be changed.

app/templates/components/toc-section.hbs
```diff
@@ -1,9 +1,10 @@
 <button
   id={{this.id}}
   class="cp-Panel-toggle level-1 list-unstyled"
-  onclick={{action (mut this.open) (not this.open)}}
+  onclick={{fn (mut this.open) (not this.open)}}
```
bertdeblock, line 4: `not` is used, but it seems `ember-truth-helpers` is not defined as a dependency.

jaredgalanis (approved):
This is a very nice cleanup, thank you! 👍

app/templates/components/search-input.hbs
jaredgalanis: good catch.

app/templates/components/search-result.hbs
jaredgalanis: I'm okay with this.

app/templates/components/toc-section.hbs
jaredgalanis: @bertdeblock it seems to be functioning as expected. We can maybe take it out in follow up.

bertdeblock: @jaredgalanis Indeed, it's pulled in by `ember-styleguide`. I can create a new PR to add the dependency? Just to make sure we don't depend on `ember-styleguide` for `ember-truth-helpers` to be included.

PR #720, opened 2020-11-19
Remove use of element event actions

bertdeblock:
[Related rule.](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-element-event-actions.md)

I think these were the only ones.

### DIFF
--- a/app/templates/components/toc-section.hbs
+++ b/app/templates/components/toc-section.hbs
@@ -1,11 +1,11 @@
 <button
   id={{this.id}}
   class="cp-Panel-toggle level-1 list-unstyled"
-  onclick={{fn (mut this.open) (not this.open)}}
   aria-label={{concat "deprecations for version " @result.since}}
   aria-expanded={{if this.open "true" "false"}}
   type="button"
   data-test-toc-list-item
+  {{on "click" (fn (mut this.open) (not this.open))}}
 >
   {{@result.since}}
 </button>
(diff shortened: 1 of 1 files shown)


app/templates/components/toc-section.hbs
```diff
@@ -15,7 +15,11 @@
     <ul aria-labelledby={{this.id}} class="level-3 list-unstyled">
       {{#each @result.contents as |content index|}}
         <li class={{if index "mt-2"}}>
-          <a class="bg-none" href="#{{id-for-deprecation content.id content.anchor}}" onclick={{this.navigateToLink}}>
+          <a
+            class="bg-none"
+            href="#{{id-for-deprecation content.id content.anchor}}"
+            {{on "click" this.navigateToLink}}
```
ijlee2, line 21: Note to myself and other maintainers:

It doesn't feel okay to include a click event on an anchor tag, and/or use `later`. Maybe, in the future, we can use a modifier to achieve what the following code does (`navigateToLink` is not the best name)?

https://github.com/ember-learn/deprecation-app/blob/e9cd33a2bdeee82c55aab63ee6e8d96474976a14/app/components/toc-section.js#L33-L42

ijlee2 (approved):
Thanks for working on allowing another recommended lint rule!

app/templates/components/toc-section.hbs
jaredgalanis: > Note to myself and other maintainers:
> 
> It doesn't feel okay to include a click event on an anchor tag, and/or use `later`. Maybe, in the future, we can use a modifier to achieve what the following code does (`navigateToLink` is not the best name)?
> 
> https://github.com/ember-learn/deprecation-app/blob/e9cd33a2bdeee82c55aab63ee6e8d96474976a14/app/components/toc-section.js#L33-L42

Yeah I know it's not great. Was trying to solve for a specific issue. Keep in mind this is all going to come out in the redesign anyway.

ijlee2: Yep, sounds good.

PR #765, opened 2021-01-26
Adds deprecation guide for v3 browser support policy

pzuraq:
Depends on https://github.com/ember-learn/ember-website/pull/745 being deployed


locks (changes requested):
marking as "request changes" while we figure out some technicalities with the browser support page link.

locks (approved):

PR #769, opened 2021-01-29
deprecation guide: window.global

NullVoxPopuli:
https://github.com/emberjs/ember.js/pull/19357/files

### DIFF
--- a/content/ember/v3/deprecate-window-global.md
+++ b/content/ember/v3/deprecate-window-global.md
@@ -1,0 +1,40 @@
+---
+id: ember-source.window-global
+title: Ember Global
+until: 4.0.0
+since: 3.26.0-beta.1
+---
+
+Accessing Ember on the global context (e.g. `window.Ember`, `globalThis.Ember`,
+or just `Ember` without importing it) is no longer supported. Migrate to
+importing Ember explicitly instead.
+
+Before:
+
+```js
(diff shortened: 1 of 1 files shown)


rwjblue (commented):
I think this should probably also be updated to mention that users needing dynamic runtime access (e.g. non-transpired access to `Ember` from the console) should use `require('ember')`.

content/ember/v3/deprecate-window-global.md
```diff
@@ -0,0 +1,40 @@
+---
+id: ember-source.window-global
+title: Ember Global
+until: 4.0.0
+since: 3.26.0-beta.1
```
mansona, line 5:
```suggestion
since: 'Upcoming Features'
```
Until the feature is actually released in a stable Ember we actually release deprecations under the "Upcoming Features" version (I know the way to do this is a bit janky but it works 😂 ) 

We could get this merged if you want as soon as this change is applied

mansona (changes requested):

jaredgalanis (commented):
@locks @NullVoxPopuli this seems to be a duplicate of https://github.com/ember-learn/deprecation-app/pull/813. Am I missing something here?

locks:
Indeed 🤔 Let's close it for now!

PR #773, opened 2021-02-02
Add property fallback deprecation

pzuraq:


locks:
@pzuraq do we not want to have an example/section on Ember Components and `@`?

pzuraq:
I'm not sure how that would be related exactly. Do you mean converting to `{{@foo}}` if you are converting from an Ember component, and are just referencing args?

I think that's basically unrelated, that is actually a fundamental change and most of the time users would probably want to just add `this` in order to maintain the existing behavior. They should only start using `{{@}}` args when they actually convert to a TO component or Glimmer component IMO.

locks (approved):

rwjblue (approved):

PR #778, opened 2021-02-02
Add deprecation guides for template transforms

pzuraq:
Adds the deprecation guides for template transformations:

- [RFC 689](https://github.com/emberjs/rfcs/blob/master/text/0689-deprecate-has-block.md)
- [RFC 690](https://github.com/emberjs/rfcs/blob/master/text/0690-deprecate-attrs-in-templates.md)
- [RFC 691](https://github.com/emberjs/rfcs/blob/master/text/0691-deprecate-class-binding-and-class-name-bindings.md)

### DIFF
--- a/content/ember/v3/has-block-and-has-block-params.md
+++ b/content/ember/v3/has-block-and-has-block-params.md
@@ -1,0 +1,72 @@
+---
+id: has-block-and-has-block-params
+title: "{{hasBlock}} and {{hasBlockParams}}"
+until: '4.0.0'
+since: 'Upcoming Features'
+---
+
+#### `{{hasBlock}}`
+
+The `{{hasBlock}}` property is true if the component was given a default block,
+and false otherwise. To transition away from it, you can use the `(has-block)`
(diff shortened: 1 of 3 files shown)


content/ember/v3/has-block-and-has-block-params.md
```diff
@@ -0,0 +1,72 @@
...
+```
+
+You may optionally pass a name to `(has-block)`, the name of the block to check.
+The name corresponding to the block that `{{hasBlock}}` represents is "default".
+Calling `(has-block)` without any arguments is equivalent to calling
+`(has-block "default")`.
+
+#### `{{hasBlockParams}}`
+
+The `{{hasBlockParams}}` property is true if the component was given a default block,
```
locks, line 43: I think you forgot to update this description when copying over.

locks (changes requested):
Small detail, otherwise looks good!

locks (approved):

rwjblue (approved):

PR #779, opened 2021-02-03
Add array observers deprecation

pzuraq:
RFC: https://github.com/emberjs/rfcs/blob/master/text/0692-deprecate-array-observers.md

### DIFF
--- a/content/ember/v3/array-observers.md
+++ b/content/ember/v3/array-observers.md
@@ -37,7 +37,11 @@ class ToStringArray {
       // else, find the new items, convert them, and add them to the array
       let newItems = innerArray.slice(changeStart, addCount);
 
-      this._content.replace(changeStart, 0, newItems.map((value) => value.toString()));
+      this._content.replace(
+        changeStart,
+        0,
+        newItems.map((value) => value.toString())
+      );
     }
 
     // Let observers/computeds know that the value has changed
@@ -50,9 +54,11 @@ class ToStringArray {
 }
 ```
 
-To move away from array observers, we could instead convert the
-wrapping to happen when the array is accessed in `objectAt`, using the `@cached`
-decorator from [tracked-toolbox](https://github.com/pzuraq/tracked-toolbox).
+To move away from array observers, we could instead convert the behavior so that
+the objects are converted into strings when the array is _accessed_ using
+`objectAt`. We can call this behavior _lazy_ wrapping, as opposed to _eager_
+wrapping which happens when the item is added to the array. We can do this using
+the using the `@cached` decorator from [tracked-toolbox](https://github.com/pzuraq/tracked-toolbox).
 
 ```js
 import { cached } from 'tracked-toolbox';
@@ -136,8 +142,10 @@ export default Component.extend({
     this._cells = A();
   },
 
-  _needsRevalidate(){
-    if (this.isDestroyed || this.isDestroying) {return;}
+  _needsRevalidate() {
+    if (this.isDestroyed || this.isDestroying) {
+      return;
+    }
     this.rerender();
   },
 
@@ -147,14 +155,14 @@ export default Component.extend({
     this.updateItems();
   },
 
-  updateItems(){
+  updateItems() {
     var rawItems = this.get('items');
 
     if (this._rawItems !== rawItems) {
       if (this._items && this._items.removeArrayObserver) {
         this._items.removeArrayObserver(this, {
           willChange: noop,
-          didChange: '_needsRevalidate'
+          didChange: '_needsRevalidate',
         });
       }
       this._rawItems = rawItems;
@@ -164,7 +172,7 @@ export default Component.extend({
       if (items && items.addArrayObserver) {
         items.addArrayObserver(this, {
           willChange: noop,
-          didChange: '_needsRevalidate'
+          didChange: '_needsRevalidate',
         });
       }
     }
@@ -181,28 +189,29 @@ export default Component.extend({
   actions: {
     scrollChange(scrollLeft, scrollTop) {
       // ...
-      if (scrollLeft !== this._scrollLeft ||
-          scrollTop !== this._scrollTop) {
+      if (scrollLeft !== this._scrollLeft || scrollTop !== this._scrollTop) {
         set(this, '_scrollLeft', scrollLeft);
         set(this, '_scrollTop', scrollTop);
         this._needsRevalidate();
       }
     },
     clientSizeChange(clientWidth, clientHeight) {
-      if (this._clientWidth !== clientWidth ||
-          this._clientHeight !== clientHeight) {
+      if (
+        this._clientWidth !== clientWidth ||
+        this._clientHeight !== clientHeight
+      ) {
         set(this, '_clientWidth', clientWidth);
         set(this, '_clientHeight', clientHeight);
         this._needsRevalidate();
       }
-    }
-  }
+    },
+  },
 });
 ```
 
-We can refactor this to update the cells themselves when they are accessed, by
-accessing them into a computed property that depends on the items array, and
-which updates the cells when it is accessed:
+We can refactor this to update the cells when they are accessed. We'll do this
+by calling `updateCells` within a computed property that depends on the items
+array:
 
 ```js
 export default Component.extend({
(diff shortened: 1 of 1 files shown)


locks (approved):

content/ember/v3/array-observers.md
```diff
@@ -0,0 +1,345 @@
+---
+id: array-observers
+title: Array Observers
+until: '4.0.0'
+since: 'Upcoming Features'
+---
+
+Array observers are a special type of observer that can be used to synchronously
+react to changes in an `EmberArray`. In general, to refactor away from them, these
+reactions need to be converted from _eager_, _synchronous_ reactions into _lazy_
```
ijlee2, line 10: I think from-to may sound better.

```suggestion
reactions need to be converted from _eager_, _synchronous_ reactions to _lazy_
```

content/ember/v3/array-observers.md
```diff
@@ -0,0 +1,345 @@
...
+reactions that occur when the array in question is _used or accessed_.
+
+For example, let's say that we had a class which wrapped an `EmberArray` and
```
ijlee2, line 13:
```suggestion
For example, let's say that we had a class that wrapped an `EmberArray` and
```

content/ember/v3/array-observers.md
```diff
@@ -0,0 +1,345 @@
...
+    innerArray.addArrayObserver(this, {
+      willChange: '_innerWillChange',
+      didChange: '_innerDidChange',
+    });
+  }
+
+  // no-op
+  _innerWillChange() {}
+
+
```
ijlee2, line 33:
```suggestion

```

content/ember/v3/array-observers.md
```diff
@@ -0,0 +1,345 @@
...
+    notifyPropertyChange(this, '[]');
+  }
+
+  objectAt(index) {
+    return this._content.objectAt(index);
+  }
+}
+```
+
+To convert this to no longer use array observers, we could instead convert the
```
ijlee2, line 55: I think it may sound a bit odd that the word convert is mentioned twice. Maybe,

```suggestion
To move away from array observers, we could instead convert the
```

content/ember/v3/array-observers.md
```diff
@@ -0,0 +1,345 @@
...
+wrapping to happen when the array is accessed in `objectAt`, using the `@cached`
```
ijlee2, line 56: Can you clarify what "convert the wrapping to happen" means? I didn't quite understand from the code below.

content/ember/v3/array-observers.md
```diff
@@ -0,0 +1,345 @@
...
+  }
+
+  objectAt(index) {
+    return this._content.objectAt(index);
+  }
+}
+```
+
+This can also be accomplished with native [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy),
+allowing your users to interact with the array using standard array syntax
```
ijlee2, line 79: I'd write 2 sentences to keep grammar simple.

```suggestion
This can also be accomplished with native [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy).
Your users can interact with the array using standard array syntax
```

content/ember/v3/array-observers.md
```diff
@@ -0,0 +1,345 @@
...
+function createToStringArray(innerArray) {
+  return new Proxy([], new ToStringArrayHandler(innerArray));
+}
+```
+
+This solution will work with autotracking in general, since users who access the
+array via `objectAt` will be accessing the tracked property. However, it will
+not integrate with computed property dependencies. If that is needed, then you
+can instead extend Ember's built-in `ArrayProxy` class, which handles forwarding
+events and dependencies itself.
```
ijlee2, line 107:
```suggestion
events and dependencies.
```

content/ember/v3/array-observers.md
```diff
@@ -0,0 +1,345 @@
...
+    return this._content.objectAt(index);
+  }
+}
+```
+
+#### Converting code that watches arrays for changes
+
+Array observers and change events can be used to watch arrays and react to
+changes in other ways as well. For instance, you may have a component like
+`ember-collection` which used array observers to trigger a rerender and
```
ijlee2, line 129:
```suggestion
`ember-collection` that used array observers to trigger a rerender and
```

content/ember/v3/array-observers.md
```diff
@@ -0,0 +1,345 @@
...
+        this._needsRevalidate();
+      }
+    }
+  }
+});
+```
+
+We can refactor this to update the cells themselves when they are accessed, by
+accessing them into a computed property that depends on the items array, and
+which updates the cells when it is accessed:
```
ijlee2, line 207: I found this sentence hard to parse. Can you break the sentence into 2 or 3 instead?

content/ember/v3/array-observers.md
```diff
@@ -0,0 +1,345 @@
...
+        set(this, '_clientHeight', clientHeight);
+        this.notifyPropertyChange('cells');
+      }
+    }
+  }
+});
+```
+
+Mutating untracked local state like this is generally ok as long as the state is
+essentially a cached representation of computed or getter is deriving in
```
ijlee2, line 250: I was confused by "getter is deriving". Was "is" meant to be included in the sentence?

content/ember/v3/array-observers.md
```diff
@@ -0,0 +1,345 @@
...
+general. It allows you to do things like compare the previous state to the
+current state during the update, and cache portions of the computation so that
+you do not need to redo all of it.
+
+It is also possible that you have some code which must run whenever the array
```
ijlee2, line 255:
```suggestion
It is also possible that you have some code that must run whenever the array
```

content/ember/v3/array-observers.md
```diff
@@ -0,0 +1,345 @@
...
+has changed, and must run eagerly. For instance, the array fragment from
+`ember-data-model-fragments` has some logic for signalling to the parent record
```
ijlee2, line 257:
```suggestion
`ember-data-model-fragments` has some logic for signaling to the parent record
```

content/ember/v3/array-observers.md
```diff
@@ -0,0 +1,345 @@
...
+      fragmentDidDirty(record, key, this);
+    } else {
+      fragmentDidReset(record, key);
+    }
+  },
+});
+```
+
+Ideally the dirty state would be converted into derived state that could read
+the array it was dependent upon, but if that's not an option or would require
```
ijlee2, line 285:
```suggestion
the array it depended on. If that's not an option or would require
```

content/ember/v3/array-observers.md
```diff
@@ -0,0 +1,345 @@
...
+      fragmentDidDirty(record, key, this);
+    } else {
+      fragmentDidReset(record, key);
+    }
+  },
+});
+```
+
+Note that this method will work for arrays and array proxies that are mutated
+directly, but will not work for array proxies which wrap other arrays and watch
```
ijlee2, line 315:
```suggestion
directly, but will not work for array proxies that wrap other arrays and watch
```

content/ember/v3/array-observers.md
```diff
@@ -0,0 +1,345 @@
...
+changes on them. In those cases, the recommendation is to refactor such that:
+
+1. Changes are always intercepted by the proxy, and can call the code
+   synchronously when they occur.
+2. The change logic is added by intercepting changes on the original array, so
+   it will occur whenever it changes.
```
ijlee2, line 321:
```suggestion
   it will occur whenever the array changes.
```

content/ember/v3/array-observers.md
```diff
@@ -0,0 +1,345 @@
...
+3. The API that must be called synchronously is instead driven by derived state.
+   For instance, in the example above, the record's dirty state could be driven
+   by the various child fragments it contains, and updated whenever the user
```
ijlee2, line 324:
```suggestion
   by the various child fragments it contains. The dirty state could be updated whenever the user
```

ijlee2 (commented):
@pzuraq It's still early morning so I might have misread and -understood the post here and there. Apologies if the suggestions aren't applicable.

I think, most of existing deprecation notes have been brief—a few paragraphs at most. I wonder if it may be better that you move what you wrote to blog.emberjs.com, as a full post, then write a summary in deprecations.emberjs.com and link to the blog post. What do you think about this idea?

(It could be that the deprecations app could instead have a collapsible-section component, so I'm not requiring that we move the post to blog.emberjs.com.)

content/ember/v3/array-observers.md
pzuraq: `_content` is accessed lazily, and so it wraps the inner values when the user actually accesses them via `objectAt`. By contrast, array observers would have wrapped eagerly, when the object was actually pushed into the array. Does that make sense?

pzuraq:
@ijlee2 merged suggestions, and updated the language in the places you pointed out were confusing. Can you re-review and let me know if it all makes sense now?

content/ember/v3/array-observers.md
```diff
@@ -0,0 +1,352 @@
...
+
+  init() {
+    this._cells = A();
+  },
+
+  cells: computed('items.[]', function() {
+    this.updateCells();
+
+    return this._cells;
+  })
```
ijlee2, line 228: Let me add the comma here.
```suggestion
  }),
```

ijlee2 (approved):
Thanks for addressing my feedback and writing up the deprecations guide for array observers. I think the post is good to go!

PR #803, opened 2021-03-01
Update Route render method deprecation version

patocallaghan:
When I initially created these docs I assumed the `Route#render`, `Route#renderTemplate` and `Route#disconnectOutlet` deprecations would ship with 3.26. They will actually ship with 3.27.


locks:
Can you please make the version `"Upcoming Features"`? when it actually releases we update to the correct version!

patocallaghan:
@locks updated 👍

ijlee2 (approved):
Thank you for keeping the Deprecation Guide up to date!

PR #805, opened 2021-03-03
Updated ember-source to v3.25

ijlee2:
## Description

I updated development dependencies to their latest version and used `ember-cli-update` to update the project from Ember 3.23 to 3.25. I recommend reviewing commits one by one.


app/components/toc-section.js
```diff
@@ -16,7 +16,7 @@ export default class TocSection extends Component {
   }
 
   get id() {
-    let dasherizedSince = this.result.since.replace(/\./g,'-');
+    let dasherizedSince = this.result.since.replace(/\./g, '-');
```
ijlee2, line 19: This component poses a technical risk because it uses classic component. When updating to Glimmer component, we want to account for 2 things:

1\. `this.result` and this.displayMobileToc` need to be written as `this.args.result` and `this.args.displayMobileToc` instead.

2\. Line 34 shows mutating the argument `@displayMobileToc`. Let's use DDAU instead.

rwwagner90 (approved):

PR #863, opened 2021-05-12
Fix metadata for deprecate-intimate-destroy-apis

locks:
The metadata for this deprecations presented two incongruences:

1. They were not explicitly marked as strings
1. "since" specified the patch version, unlike other deprecations


mansona (approved):
looks good to me 👍  we should probably think of having a test that makes sure this would fail in future PRs as we might regress 🤷

rwwagner90 (approved):

PR #932, opened 2021-08-20
Implicit injection deprecations for v4.0.0

snewcomer:
Per discussion at https://github.com/emberjs/ember.js/pull/19680/files#r692596832, add a deprecation guide entry for implicit injection APIs deprecation in 4.0.

### DIFF
--- a/content/ember/v4/deprecate-implicit-injection.md
+++ b/content/ember/v4/deprecate-implicit-injection.md
@@ -6,3 +6,40 @@ since: '4.0.0'
 ---
 
 Implicit injections have been [deprecated](https://deprecations.emberjs.com/v3.x#toc_implicit-injections) since Ember v3.26.0.  As of v4.0.0, implicit injections do nothing and should be removed based on suggestions in the original deprecation.
+
+Before:
+
+```js
+export default {
+  initialize(app) {
+    app.inject('route', 'store', 'service:store');
+  }
+}
+```
+
+```js
+import { Route } from '@ember/routing/route';
+
+export default class ApplicationRoute extends Route {
+  model() {
+    return this.store.findQuery('user', 123);
+  }
+}
+```
+
+After:
+
+```js
+import { Route } from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+
+export default class ApplicationRoute extends Route {
+  @service store;
+
+  model() {
+    return this.store.findQuery('user', 123);
+  }
+}
+```
+
+For a more detailed explanation with additional examples, see the old deprecation [guides](https://deprecations.emberjs.com/v3.x#toc_implicit-injections).
--- a/content/ember/v4/deprecate-implicit-injection.md
+++ b/content/ember/v4/deprecate-implicit-injection.md
@@ -1,0 +1,8 @@
+---
+id: implicit-injections
+title: "Implicit Injections"
+until: '5.0.0'
+since: '4.0.0'
+---
+
+Implicit injections have been [deprecated](https://deprecations.emberjs.com/v3.x#toc_implicit-injections) since Ember v3.26.0.  As of v4.0.0, implicit injections do nothing and should be removed based on suggestions in the original deprecation.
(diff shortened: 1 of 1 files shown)


content/ember/v4/deprecate-implicit-injection.md
```diff
@@ -0,0 +1,8 @@
+---
+id: implicit-injections
+title: "Implicit Injections"
+until: '5.0.0'
+since: '4.0.0'
+---
+
+Implicit injections have been [deprecated](https://deprecations.emberjs.com/v3.x#toc_implicit-injections) since Ember v3.26.0.  As of v4.0.0, implicit injections do nothing and should be removed based on suggestions in the original deprecation.
```
mixonic, line 8: Can we provide an example of what to remove here?

I think this description is needlessly confusion for someone stumbling into this deprecation without any context. A certain part of the functionality was deprecated in 3.26, but that is *not* what is deprecated in 4.0 for removal in 5.0. The behavior deprecated in 3.26 is changed in 4.0 already.

Providing a link back to the old deprecation guide is useful context, but the message here has to be something other than "go read the old guide", as the old guide was targeted for apps upgrading to 4.0 and not 5.0.

mixonic (changes requested):

content/ember/v4/deprecate-implicit-injection.md
snewcomer: This is true.  All this most recent PR did was move the deprecation from the object receiving the injection to the inject method itself.  The only difference being the method just doesn't work now.

I've done a few of these upgrades in addons and apps and the solution has varied widely (some solutions not included in the original deprecation guide).  I want to provide good direction so if we think providing an extensive guide similar to the last one is a good idea, then happy to do it.  Effectively, the old guides recommend moving to either `@service` or `getOwner`, both of which seem to be a common fix most of the time.  

If, for example we just include a simple example moving from `owner.inject` to `@service`, then we might not be giving a reader the whole picture of what options you might have to take.

mixonic: This is better after the update. I think the theory in 4.0, however, is that the inject call can be removed because it is always a no-op, right?

Telling people to migrate a usage to explicit injections isn't needed, because those were already deprecated in 3.26 and apps should not have them?

Is the guide as simple as being explicit that "you have code that looks like [example], in 4.0 you should delete calls to the injection API since they no longer do anything. See the 3.26 deprecation for more" ?

snewcomer: One minor consideration is there are probably cases where ppl will upgrade from <3.26 (thus didn't hit the deprecation) and those that did hit the deprecation.  I like your suggestions though.  I feel like we might be hitting a middle ground with the current state of the PR - those that are seeing this deprecation for the first time and those that haven't.

jaredgalanis: @snewcomer @mixonic just checking in on this. It does look like maybe the after should just import `service` w/o `inject` given it's availability in 4.1.

bartocc:
Hey folks, what's the status of this PR? it would be great to have this deprecation added to the guides

content/ember/v4/deprecate-implicit-injection.md
locks: @jaredgalanis I don't think that is wise since the deprecation is prior to 4.1.

snewcomer:
@bartocc agreed. Been close to 9 months now and something as simple as this would have benefitted from being merged (even with further considerations to be made) for folks rather than sitting in limbo 😢.  If I could merge, I would have.

bartocc:
Awesome 🎉

Would it be desirable/possible to also update the section `Deprecations added in 4.0` in the [Ember 4.0 released](https://blog.emberjs.com/ember-4-0-released/) post?

locks:
Send a PR over and we can review!

bartocc:
> Send a PR over and we can review!

I'll do that 👍

Issue #893, opened 2021-07-03
Missing deprecation guides


locks:
- [ ] Ember.Component.reopen `ember.component.reopen`[Deprecation Guide]() [Deprecation Invocation Site](https://github.com/emberjs/ember.js/blob/9609b8f30e0cdbc2235063f571af619f4362f73b/packages/@ember/-internals/glimmer/lib/component.ts#L1156)
- [ ] Passing tagName to LinkTo `ember.link-to.tag-name` [Deprecation Guide]() [Deprecation Invocation Site](https://github.com/emberjs/ember.js/blob/57907a4d7d91aa58b3b3ac1258e639ce221f9492/packages/@ember/-internals/glimmer/lib/components/link-to.ts#L432)
- [ ] Passing disabledWhen to LinkTo `ember.link-to.disabled-when` [Deprecation Guide]() [Deprecation Invocation Site](https://github.com/emberjs/ember.js/blob/57907a4d7d91aa58b3b3ac1258e639ce221f9492/packages/@ember/-internals/glimmer/lib/components/link-to.ts#L627) + 1 more in the same file
- [ ] Old deprecate method imports `old-deprecate-method-paths` [Deprecation Guide]() [Deprecation Invocation Site](https://github.com/emberjs/ember.js/blob/22bfcfdac0aeefcf333fb2d6697772934201b43b/packages/@ember/application/de precations.ts#L8) + 1 other location in the same file
- [ ] Deprecate without `for` `ember-source-deprecation-without-for` [Deprecation Invocation Site](https://github.com/emberjs/ember.js/blob/4984045459bc9f5f34f1ae8b51fc6522fb72cbf4/packages/@ember/debug/lib/deprecate.ts#L211) 
- [ ] Deprecate without `since` `ember-source-deprecation-without-since` [Deprecation Invocation Site](https://github.com/emberjs/ember.js/blob/4984045459bc9f5f34f1ae8b51fc6522fb72cbf4/packages/@ember/debug/lib/deprecate.ts#L224) 
- [ ] EXTEND_PROTOTYPES `ember-env.old-extend-prototypes` [Deprecation Guide]() [Deprecation Invocation Site](https://github.com/emberjs/ember.js/blob/566af4941eb7ecaa4679e295ecfc38263d5ee1fe/packages/ember/index.js#L175)
[...]

PR #955, opened 2021-09-08
Deprecation guide Ember.link-to @disabled-when

Glarregle:
Progress #893

### DIFF
--- a/content/ember/v4/ember-link-to-disabled-when.md
+++ b/content/ember/v4/ember-link-to-disabled-when.md
@@ -1,0 +1,23 @@
+---
+id: ember.link-to.disabled-when
+title: LinkTo @disabled-when argument
+until: '5.0.0'
+since: '4.0.0'
+---
+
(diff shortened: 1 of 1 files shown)


app/routes/show.js
```diff
@@ -4,6 +4,7 @@ import RSVP from 'rsvp';
 
 export default class ShowRoute extends Route {
   @service resultProcessor;
+  @service store;
```
locks, line 7: Could you make a separate PR with this change? I know it's only a small line, but it'll help keep commits organized. Appreciated :)

locks (approved):
Thanks for the help!

bertdeblock:
Sorry to comment on a merged PR, but it seems the `until` value used here is not the same as for the actual deprecation: https://github.com/emberjs/ember.js/blob/57907a4d7d91aa58b3b3ac1258e639ce221f9492/packages/@ember/-internals/glimmer/lib/components/link-to.ts#L635.

PR #956, opened 2021-09-08
Deprecation guide Old deprecate Method Path

Glarregle:
Progress #893


locks:
The team had incorrectly identified the `since` and `until` versions of the deprecation. Could you move this content to https://github.com/ember-learn/deprecation-app/blob/main/content/ember/v3/old-deprecate-method-paths.md ? Thank you for the help!

PR #964, opened 2021-09-15
testing percy, do not merge

jenweber:
I'm seeing a problem with Percy in another project and I'd like to see if this one is ok.


jenweber:
There was a Percy service outage yesterday, and this PR helped confirm it. The status board was green but now there is some info on the status page about the incident.

PR #965, opened 2021-09-17
Update deprecation until for deprecated run loop

jenweber:
This was incorrect - it is removed in `master` so it's "until" v4.


jenweber:
Self merging since this is a tiny fix I verified in ember source.

PR #967, opened 2021-09-17
Update deprecation until for registerPlugin

jenweber:


jenweber:
Self merging since this is a tiny fix I verified in ember source.

PR #973, opened 2021-09-24
Update ember-env-old-extend-prototypes.md

fannyvanrell:
✏️ Update typo


locks:
Hi! Do you want to get this change in and then send another patch with updated instructions?

bartocc:
Hi @locks. Nope, I believe we'll do everything in this PR, unless you advice otherwise

locks:
Go ahead 👍

fannyvanrell:
Hi @locks, we've added examples in the deprecation description, could you add your review?

PR #999, opened 2021-10-27
Improving doc for deprecation "deprecated-run-loop-and-computed-dot-access"

leonardoraele:
- **Source:** https://github.com/emberjs/ember.js/pull/19653/files
- **Note:** The properties are listed per-line for ease of writing, but they will be displayed inline in the rendered document because markdown.


locks (approved):
Thanks for the help :)

PR #1129, opened 2022-04-28
Update `template-compiler.registerPlugin` guide

bertdeblock:


vmoglan (approved):
Before our discussion I thought the error was something coming from our app, but now it completely makes sense. Thanks a lot for updating!

locks (approved):
Thanks @bertdeblock :)
I moved the note to before the code sample so it's more likely for people to see it.

PR #1133, opened 2022-05-16
Remove unnecessary single quote

camerondubas:
Just a very minor typo I noticed while reading the v3 deprecations page. Cheers!


MinThaMie (approved):
Thank you very much for noticing and fixing :)

PR #1179, opened 2022-08-01
Correctly apply deprecate-router-events changes

wagenet:
The change should be applied to the Router, not a Route.

--

Should be squashed before merging.

### DIFF
--- a/content/ember/v3/router-events.md
+++ b/content/ember/v3/router-events.md
@@ -41,7 +41,7 @@ export default Router.extend({
 To:
 
 ```js
-import Route from '@ember/routing/route';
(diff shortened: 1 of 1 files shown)


content/ember/v3/router-events.md
```diff
@@ -41,7 +41,7 @@ export default Router.extend({
 To:
 
 ```js
-import Route from '@ember/routing/route';
+import Router from '@ember/routing/router';
 import { inject as service } from '@ember/service';
 
 export default Route.extend({
```
wagenet, line 47:
```suggestion
export default Router.extend({
```

jenweber (approved):
Thank you!

PR #1192, opened 2022-09-01
Add deprecation guide for array prototype extensions for V5

smilland:
[RFC: Deprecate array prototype extensions](https://github.com/emberjs/rfcs/pull/848)

[Rendered content](https://github.com/smilland/deprecation-app/blob/deprecate-array-prototype-extensions/content/ember/v5/deprecate-array-prototype-extensions.md)

It's ready for **_review_**, though only can be merge after the RFC is approved.


content/ember/v5/deprecate-array-prototype-extensions.md
```diff
@@ -0,0 +1,1265 @@
...
+#### `compact`
+
+Before:
+```js
+someArray.compact();
+```
+
+After:
+```js
+someArray.filter(val => val !=== undefined && val !== null);
```
tniezurawski, line 40: I think there's one `=` too much here:

```suggestion
someArray.filter(val => val !== undefined && val !== null);
```

Is it still recommended these days to check `undefined` through `typeof` or not? 🤔 

```suggestion
someArray.filter(val => typeof val !== 'undefined' && val !== null);
```

smilland:
thank you so much for reviewing @tniezurawski !

content/ember/v5/deprecate-array-prototype-extensions.md
```diff
@@ -0,0 +1,1265 @@
...
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+
+export default class SampleComponent extends Component {
+  @tracked abc = ['x', 'y', 'z', 'x'];
+
+  @action
+  pushObject(value) {
+    this.abc = [...this.abc, value];
```
elgordino, line 730: I'm a little confused by the difference between this example and the `pushObject` example above which uses `TrackedArray`. Is it a requirement to assign to `this.abc` when using `@tracked` but not when using `TrackedArray`? Or is the different method shown to indicate that both approaches are valid?

content/ember/v5/deprecate-array-prototype-extensions.md
```diff
@@ -0,0 +1,1265 @@
...
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { TrackedArray } from 'tracked-built-ins';
+
+export default class SampleComponent extends Component {
+  abc = new TrackedArray(['x', 'y', 'z', 'x']);
+
+  @action
+  pushObjects(values) {
+    this.abc.splice(this.abc.length, 0, ...values);
```
elgordino, line 761: Would `this.abc.push(...values)` work here? It would be simpler and match the first `after` example for `pushObject` above?

content/ember/v5/deprecate-array-prototype-extensions.md
```diff
@@ -0,0 +1,1265 @@
...
+After:
+```js
+const someArray = [{ food: 'apple', isFruit: true }, { food: 'beans', isFruit: false }];
+[...someArray].sort((a, b) => {
+  return a.food?.localeCompare(b.food)
+    ? a.food?.localeCompare(b.food)
+    : a.isFruit - b.isFruit;
+}); // [{ food: 'apple', isFruit: true }, { food: 'beans', isFruit: false }]
+```
+#### `toArray`
```
simonihmig, line 230: Is this example even needed? I think you would use `toArray` an an array-like object like `A()`, to turn it into a native array. But here we are only dealing with native arrays, so what's the point of `toArray()` usage then?

content/ember/v5/deprecate-array-prototype-extensions.md
```diff
@@ -0,0 +1,1265 @@
...
+Before:
+```js
+const someArray = [1, 2, 3, undefined, 3];
+someArray.uniq(); // [1, 2, 3, undefined]
+```
+
+After:
+```js
+const someArray = ['a', 'a', 'b', 'b']
+[...new Set(someArray)] // ['a', 'b']
```
simonihmig, line 255: This example should match the one above, so rather like:

```suggestion
const someArray = [1, 2, 3, undefined, 3];
[...new Set(someArray)] // [1, 2, 3, undefined]
```

content/ember/v5/deprecate-array-prototype-extensions.md
```diff
@@ -0,0 +1,1265 @@
+---
+id: deprecate-array-prototype-extensions
+title: "Deprecate array prototype extensions"
+until: '6.0.0'
+since: '5.0.0'
```
simonihmig, line 5: Should we update this to the current one on `main`, assuming we get the deprecation implemented in that version?

```suggestion
since: '5.10.0'
```

content/ember/v5/deprecate-array-prototype-extensions.md
```diff
@@ -0,0 +1,1265 @@
...
+
+After
+```js
+// TrackedArray
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { TrackedArray } from 'tracked-built-ins';
+
+export default class SampleComponent extends Component {
+  abc = new TrackedArray();
```
simonihmig, line 392:
```suggestion
  abc = new TrackedArray(['x', 'y', 'z', 'x']);
```

content/ember/v5/deprecate-array-prototype-extensions.md
simonihmig: yeah, `@tracked` will only trigger the reactivity when a new value gets assigned. When the existing native array is internally mutated, like calling `.push()`, then this will not be reactive. In opposite to `TrackedArray`, which does this kind of deep tracking. So I think the example is right.

content/ember/v5/deprecate-array-prototype-extensions.md
simonihmig: yeah, i think so!

content/ember/v5/deprecate-array-prototype-extensions.md
```diff
@@ -0,0 +1,1265 @@
...
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+
+export default class SampleComponent extends Component {
+  @tracked abc = ['x', 'y', 'z', 'x'];
+
+  @action
+  removeAt(start, len) {
+    this.abc.splice(start, len);
+    this.abc = [...this.abc];
```
simonihmig, line 827: I think this would also work, and be a bit simpler?

```suggestion
    this.abc.splice(start, len);
    this.abc = this.abc;
```

content/ember/v5/deprecate-array-prototype-extensions.md
```diff
@@ -0,0 +1,1265 @@
...
+  @action
+  removeObject(value) {
+    let loc = this.abc.length || 0;
+    while (--loc >= 0) {
+      let curValue = this.abc.at(loc);
+
+      if (curValue === value) {
+        this.abc.splice(loc, 1);
+      }
+    }
```
simonihmig, line 865: Wouldn't a simple filer call work here?

```suggestion
    this.abc.filter(item => item !== value);
```

same for the related examples below...

content/ember/v5/deprecate-array-prototype-extensions.md
```diff
@@ -0,0 +1,1265 @@
...
+  constructor(name) {
+    this.name = name;
+  }
+
+  greet(prefix = 'Hello') {
+    return `${prefix} ${this.name}`;
+  }
+}
+
+[new Person('Tom'), new Person('Joe')].map(person => person['greet']?.('Hi')); // ['Hi Tom', 'Hi Joe']
```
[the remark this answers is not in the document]
ef4: `invoke` returns a value like `map`, so even though this example doesn't motivate using the return value I think it's correct to tell people to use `map`.

content/ember/v5/deprecate-array-prototype-extensions.md
```diff
@@ -0,0 +1,1265 @@
...
+Before:
+```js
+const someArray = [{ food: 'apple', isFruit: true }, { food: 'beans', isFruit: false }];
+someArray.sortBy('food', 'isFruit'); // [{ food: 'apple', isFruit: true }, { food: 'beans', isFruit: false }]
+```
+
+After:
+```js
+const someArray = [{ food: 'apple', isFruit: true }, { food: 'beans', isFruit: false }];
+[...someArray].sort((a, b) => {
```
ef4, line 224: `toSorted` is new and it's a non-mutating version of `sort`:

```suggestion
someArray.toSorted((a, b) => {
```

content/ember/v5/deprecate-array-prototype-extensions.md
```diff
@@ -0,0 +1,1265 @@
...
+  return a.food?.localCompare(b.food)
+    ? a.food?.localCompare(b.food)
+    : a.isFruit - b.isFruit;
+}); // [{ food: 'apple', isFruit: true }, { food: 'beans', isFruit: false }]
+```
```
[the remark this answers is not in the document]
ef4: We could also give people a replacement function they can paste into their own code. This is adapted from ember's own implementation:

```ts
import { compare } from '@ember/utils';
function sortBy<T>(array: T[], ...sortKeys: string[]): T[] {
  return this.toSorted((a: T, b: T) => {
    for (let i = 0; i < sortKeys.length; i++) {
      let key = sortKeys[i];
      let propA = a[key];
      let propB = b[key];
      let compareValue = compare(propA, propB);
      if (compareValue) {
        return compareValue;
      }
    }
    return 0;
  });
}
```

ef4 (commented):
Two other general notes that would be nice to include in this guide:

1. When the array prototype extensions are off, passing a native array to `import { A } from '@ember/array'` mutates the native array to re-add the prototype extensions. This can cause surprising behavior. (If somebody is excited to also deprecate `A` it's worth doing too.)
2. When the array prototype extensions are off, if you take an array-like value from ember-data and pass it to `A`, it breaks.

Other than these suggestions, this looks good to go. 👍

ef4:
This deprecation is activated in ember beta 5.10, so it would be great to get this landed. I think only a small amount of minor fixes are needed here to get this green and mergeable.

content/ember/v5/deprecate-array-prototype-extensions.md
```diff
@@ -0,0 +1,1265 @@
...
+
+## Observable Properties
+`firstObject`, `lastObject` are observable properties. Changing directly from `firstObject` to `at(0)` or `[0]` might cause issues that the properties are no longer reactive.
+
+### Used in template
+If the `firstObject` and `lastObject` are used in a template, you can convert to use `get` helper safely as `get` helper handles the reactivity already.
+
+Before
+```hbs
```
jaredgalanis, line 352:
```suggestion
```handlebars
```

content/ember/v5/deprecate-array-prototype-extensions.md
```diff
@@ -0,0 +1,1265 @@
...
+<Foo @bar={{@list.firstObject.name}} />
+```
+
+After
+```hbs
```
jaredgalanis, line 357:
```suggestion
```handlebars
```

content/ember/v5/deprecate-array-prototype-extensions.md
jaredgalanis: > 

This should fix the broken test.

content/ember/v5/deprecate-array-prototype-extensions.md
```diff
@@ -0,0 +1,1265 @@
...
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { TrackedArray } from 'tracked-built-ins';
+
+export default class SampleComponent extends Component {
+  abc = new TrackedArray(['x', 'y', 'z', 'x']);
+
+  @action
+  pushObjects(values) {
+    this.abc.splice(this.abc.length, 0, ...values);
```
ef4, line 761:
```suggestion
    this.abc.push(...values);
```

ef4 (approved):

ef4:
I don't have permissions to do the Percy approval but it passes a visual inspection. Merging, because this deprecation is already released and we need a non-broken link.

PR #1238, opened 2022-12-08
Add deprecation note for `autotracking.mutation-after-consumption`

chriskrycho:
This was changed in the Glimmer VM upgrade which landed as part of Ember 4.0, but there was no user-facing information about what the issue is or how to resolve it, and we previously had no infrastructure for these kinds of changes. This resolves both:

- Add support for a new `'Glimmer Internals'` category (f510f4a)
- Add the write-up itself (5909a6d)

However, in the process of making that change, I noticed that we an easy win in terms of performance and maintainability, switching away from an Ember Concurrency `Task` to just using a simple function to compute the results needed for the `ember` and `show` routes (6fadd0f). I also made some other small modernization (29f5ed9) and performance (064ae09) improvements.

https://user-images.githubusercontent.com/2403023/206567511-919e15cc-c49f-441b-8eb6-d45009f529d8.mp4


chriskrycho:
I'd love folks on learning to come back and iterate on this later, but I'm going to go ahead and merge it: this was a significant problem for LinkedIn apps looking to upgrade, because it was *invisible*.

Issue #1239, opened 2022-12-12
[BUG] Mobile  Deprecation version section is too long


arafatiqbal:
The issue on screen size 390x844 (iPhone 12 pro): 
![image](https://user-images.githubusercontent.com/70502771/206941269-184a788c-2d4d-4ce1-bb85-674f7bba93b8.png)

Maybe making another column on the right, and having only 3 version per column? 
Something like this:
![image](https://user-images.githubusercontent.com/70502771/206941469-1f5acea1-5b03-4b26-a75c-e58de38941e2.png)
I know this would only fix it up until ember v7, but I doubt that will happen within the next couple years, so until that happens this is the easiest solution I could think of.

arafatiqbal:
Or I guess, we could increase the height as well?

PR #1240, opened 2022-12-12
Mobile Deprecation version section is too long

arafatiqbal:
#1239 

Before: 
Screen Shot 2022-12-11 at 7 34 30 PM

After:
Screen Shot 2022-12-11 at 7 34 54 PM

There is now less empty space on the right side, and the proportions just look a bit better.

### DIFF
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -581,7 +581,7 @@ p a:hover {
         width: 100%;
 
         .ember-version-graphic {
-          $height: 132px;
+          $height: 161px;
           width: $height;
           height: $height;
 
@@ -592,6 +592,14 @@ p a:hover {
 
         .links {
           border-top: 0;
+       }
+        .row{
+          display: flex;
+          flex-direction: row;
+        }
+        .column{
+          display: flex;
+          flex-direction: column;
         }
       }
     }
--- a/app/templates/index.hbs
+++ b/app/templates/index.hbs
@@ -36,18 +36,24 @@
     <li class="item list-unstyled">
       <EmberVersionGraphic @mascot="ember" />
       <ul class="links">
-        <li class="list-unstyled" data-test-ember-1-link>
+        <div class="row">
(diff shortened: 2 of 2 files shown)


app/styles/app.scss
```diff
@@ -581,7 +581,7 @@ p a:hover {
         width: 100%;
 
         .ember-version-graphic {
-          $height: 132px;
+          $height: 161px;
```
arafatiqbal, line 584: This makes the ember-version-graphic a tad bit larger, so 3 version numbers could fit per column. I also think it looks a bit better with it larger

app/templates/index.hbs
```diff
@@ -36,18 +36,24 @@
     <li class="item list-unstyled">
       <EmberVersionGraphic @mascot="ember" />
       <ul class="links">
-        <li class="list-unstyled" data-test-ember-1-link>
+        <div class="row">
```
jenweber, line 39: I think that a `ul` must have `li` as direct children, without divs in between, based on the "Permitted content" section of the `ul` docs: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul#technical_summary

One way to achieve this layout without the divs, while still keeping this one continuous list, is something like https://codepen.io/srikarg/pen/kxwQXV

There are lots of options for how to write this CSS. Another example is with flexbox: https://stackoverflow.com/questions/42613350/how-can-i-create-multi-columns-from-a-single-unordered-list

jenweber (changes requested):
I like your design! It fixes what looks like a visual bug. I think the HTML markup needs some fixes before this can be merged, to ensure that our HTML is valid and accessible. Let me know if you have any questions or want help. Thank you for your work on this!

app/templates/index.hbs
arafatiqbal: Got it thanks! I'll push up a few commits to address this when I get a chance

Issue #1241, opened 2022-12-12
Add markdown linting with remark-lint


arafatiqbal:
I noticed remark-lint was used in the guides-source repo, figured it wouldn't hurt to add the md linting here as well

PR #1242, opened 2022-12-12
Add remark-lint for markdown files

arafatiqbal:
#1241


jenweber (approved):
Thank you! This would be nice to have for consistency!

If you or anyone else is interested in moving this forward, I think the next steps are to add a config, an ignore list, and add the check to our CI tests.

An example remark config is in https://github.com/ember-learn/guides-source/blob/master/.remarkrc.js

PR #1260, opened 2023-01-13
Fix duplicate v4.10 entries

bertdeblock:
Fixes:

Screenshot 2023-01-13 at 13 24 59


wagenet (changes requested):
I believe this should be 4.10.0. We should change the other spots.

bertdeblock:
This change is consistent with all v1, v2 and v3 deprecations. AFAICT, this PR was correct and I don't understand why it has been closed.

Issue #1276, opened 2023-01-30
deprecation: using / for division outside of calc()


geneukum:
When I build the app, Dart Sass logs the warning:

```
Deprecation Warning: Using / for division outside of calc() is deprecated and will be removed in Dart Sass 2.0.0.

Recommendation: math.div($base-spacing, 2) or calc($base-spacing / 2)

More info and automated migrator: https://sass-lang.com/d/slash-div

   ╷
19 │ $small-spacing: $base-spacing / 2;
```

We should fix this issue using the provided recommendation in order to clean up the build logs and make migration to dart sass 2.0.0 easier in the future.

locks:
Fixed by linked PR. Thanks!

PR #1277, opened 2023-01-30
_search-input: sass: avoid using slash as div

geneukum:
Sass currently treats / as a division operation in some contexts and a separator in others. This makes it difficult for Sass users to tell what any given / will mean, and makes it hard to work with new CSS features that use / as a separator.

As such, Sass has deprecated its usage, and `dart-sass` (which we use in this project) flags it at build time.

To fix this issue, let's prefer `math.div` instead.

See: https://sass-lang.com/documentation/breaking-changes/slash-div for more details.

This fixes issue #1276.

### DIFF
--- a/app/styles/_search-input.scss
+++ b/app/styles/_search-input.scss
@@ -1,3 +1,4 @@
+@use "sass:math";
 
 // Typography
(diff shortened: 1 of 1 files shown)


jenweber (approved):
Thank you!

Issue #1272, opened 2023-01-29
Upgrade this app to Ember 4


jenweber:
This app is on Ember 3. We should upgrade to Ember 4!

Instructions: https://guides.emberjs.com/release/upgrading/

Whoever picks up this issue could upgrade to v4.0 or a later version. It's up to you!

I am marking this good first issue, but it is only a good first issue for someone who has experience upgrading another Ember app.

PR #1278, opened 2023-01-30
package: remove `ember-collapsible-panel` dep

geneukum:
At build-time, a warning is logged:
```
Usages of the Ember Global may be caused by an outdated ember-cli-babel
dependency. The following steps may help:
* Upgrade the following addons to the latest version:
  ...
  * ember-collapsible-panel
```

We *do* depend on the `ember-collapsible-panel` package, but do not appear to actually use it.  For reference, it looks like it was added in `Commit: aecc6b5d`. and removed in `Commit: dda4d0e5`.

As such, we should just be able to remove the dependency instead of trying to upgrade it.

This should help along the way to achieving:
https://github.com/ember-learn/deprecation-app/issues/1272.

### DIFF
--- a/app/templates/components/toc-section.hbs
+++ b/app/templates/components/toc-section.hbs
@@ -1,6 +1,6 @@
 <button
   id={{this.id}}
-  class="cp-Panel-toggle level-1 list-unstyled"
+  class="level-1 list-unstyled"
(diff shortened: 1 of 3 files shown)


jenweber (approved):
I found one lingering class hanging around that had the cp prefix, and it didn't seem to do anything. Looks good to me!

PR #1279, opened 2023-01-30
package: update `ember-styleguide`, `ember-showdown-prism`, `ember-auto-import` deps

geneukum:
At build time, a warning is logged:
```
Usages of the Ember Global may be caused by an outdated ember-cli-babel
dependency. The following steps may help:

* Upgrade the following addons to the latest version:
  * ember-styleguide
```

As such, let's bump to the newest version of this package. See https://github.com/ember-learn/ember-styleguide/releases for the package's release notes.

This should help along the way to achieving: #1272.


geneukum:
Ah, sorry about that!

I'll take a look tonight and see if I can get a working change set out for your review.

geneukum:
Hi @jenweber, I think that the linter issues are coming from `ember deprecation-lint`. At a high level, it looks like there are really two separate problems here:
- We (and our dependencies) depend on different incompatible versions of `@ember/render-modifiers`
- Our `deprecation-lint` config setting for the allowed versions of `@ember-concurrency` does not include the version that we and `ember-styleguide` both want to use.

Digging into the first issue a bit, the dependency tree looks like:
```
deprecation-app

├── @ember/render-modifiers@1.0.2
├─┬ ember-showdown-prism
│ └─┬ ember-prism
│   └── @ember/render-modifiers@1.0.2
└─┬ ember-styleguide
  └── @ember/render-modifiers@2.0.5
```
Interestingly, we do depend on render-modifiers, but don't appear to actually use it from what I can tell, so I submitted #1283 to remove that. For `ember-showdown-prism`, I went ahead and updated that in this pull request since it needs to be updated in lockstep with `ember-styleguide`.

For the second issue, since we and `ember-styleguide` want to use the same version, I think that we should just be able to remove that `allowedVersion` setting in the `config/dependency-lint.js` file to resolve the linter error.

I _think_ that once https://github.com/ember-learn/deprecation-app/pull/1283 has been merged, this should lint successfully (or at least it did when I merged it locally).

Please let me know if there's anything else that I can do here! 😄

jenweber:
I updated this branch with `main` after merging #1283. The tests are still unhappy, but this time we get a different error output:

> this version of ember-auto-import requires the app to have a dependency on webpack 5

I think possibly this can be solved by installing webpack. It looks like they are still on v5.

`npm install --save-dev webpack`

geneukum:
Ah, darn! I'm not sure why I'm not seeing that locally... I'll try your suggestion and see if that makes the CI happy. Sorry for the inconvenience here.

geneukum:
Thanks for the suggestion, that seems to have done the trick! 

Please let me know if there's anything else that we should do here. 😃

jenweber (approved):
Everything looks great in the click test, thank you!

PR #1284, opened 2023-01-31
Add ember cli deprecation workflow

arafatiqbal:
I figured this is a great addition to have to help keep track of the deprecations in the app. 

Fixing this list of deprecations will/ might help with the upgrade #1272

### DIFF
--- a/config/deprecation-workflow.js
+++ b/config/deprecation-workflow.js
@@ -1,0 +1,11 @@
+/* eslint-disable */
+
+window.deprecationWorkflow = window.deprecationWorkflow || {};
+window.deprecationWorkflow.config = {
+  workflow: [
+    { handler: 'silence', matchId: 'ember-global' },
+    { handler: 'silence', matchId: 'manager-capabilities.modifiers-3-13' },
+    { handler: 'silence', matchId: 'ember.built-in-components.import' },
+    { handler: 'silence', matchId: 'this-property-fallback' },
+  ],
+};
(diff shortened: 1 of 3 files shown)


config/deprecation-workflow.js
```diff
@@ -0,0 +1,11 @@
...
+
+window.deprecationWorkflow = window.deprecationWorkflow || {};
+window.deprecationWorkflow.config = {
+  workflow: [
+    { handler: 'silence', matchId: 'ember-global' },
+    { handler: 'silence', matchId: 'manager-capabilities.modifiers-3-13' },
+    { handler: 'silence', matchId: 'ember.built-in-components.import' },
+    { handler: 'silence', matchId: 'this-property-fallback' },
+  ],
+};
```
arafatiqbal, line 11: You have to manually update this config which kind of sucks. 
1. Run the tests suite or navigate through the app
2. Run `deprecationWorkflow.flushDeprecations()` in the console
3. Copy the output, and paste it in here.

If you fix one of the deprecations, re-run the above, and update this file

jenweber:
Yeah, this addon is super helpful when there are a lot of deprecations!

Before merging, let's check in with @geneukum who has being doing a bunch of the upgrading. What do you think about adding this? Is there any particular config that would be helpful?

geneukum:
Yeah, this looks like a super useful addition to me as well!

I'd be curious which (if any) of the deprecations show up after we merge https://github.com/ember-learn/deprecation-app/pull/1287 and https://github.com/ember-learn/deprecation-app/pull/1279 though.

I guess that we may need to just be careful about ordering to make sure that we're not either silencing deprecations that don't need to be silenced or leaving around silences that have been resolved elsewhere?

PR #1285, opened 2023-01-31
package: fix `LinkComponent` deprecation

geneukum:
Right now, we import Ember.LinkComponent to define custom click event handling.

Using Ember.LinkComponent or importing from 'LinkComponent' has been deprecated, and it is recommended to install the
`@ember/legacy-built-in-components` addon and use `import { LinkComponent } from '@ember/legacy-built-in-components';` instead [deprecation id: ember.built-in-components.import], so let's go ahead and do so.

This should help with achieving #1272.


jenweber (approved):
Thank you! I click tested and link clicks seem to work how I expect.

PR #1286, opened 2023-02-01
avoid depending on implicit store injection in routes

geneukum:
Right now, the `ember` and `show` routes both depend on implicit injection of the store service. Ember is moving away from this paradigm, so let's migrate away from this in favor of implicit injection via the `inject` decorator.

This should help make incremental progress toward #1272.


jenweber (approved):
Thank you!

PR #1287, opened 2023-02-03
tests: use `this` in property path in templates

geneukum:
When running the `id-for-deprecation-test` suite, there are several instances of the [this-property-fallback] deprecation logged to the console.

Let's make sure that we always use `this.` in the property path of our inline templates when appropriate.

This should help with #1272.


jenweber (approved):

jenweber:
Thank you!

PR #1261, opened 2023-01-13
Make individual pages for deprecations

jenweber:
Closes https://github.com/ember-learn/deprecation-app/issues/1256

Goal:
URLS like like `https://deprecations.emberjs.com/id/some-deprecation` bring you to a guide about an individual deprecation

Progress so far:

Screen Shot 2023-01-29 at 4 13 45 PM

PR #1339, opened 2023-08-30
Add deprecations from Ember Data 5.x so far

Baltazore:


locks (changes requested):
can't merge since the changes contain an unreleased deprecation

Baltazore:
@locks do you want me to split PR into 2 separate (one with 5.0 deprecation, another with 5.3 deprecations) ? Or this one can just wait until EData 5.3 would be released ?

runspired:
@locks this PR is happening now because @kategengler asked us to PR it and get url's prior to release. Sounds like there's a bit of a mixup somewhere in what's expected, we've never used urls for EmberData deprecations in the past but happy to use them going forward. From what I can tell they are predictable off of the file names in this PR so as long as there are no concerns about the file names I'll update the URLs in-advance in EmberData. Pattern: `https://deprecations.emberjs.com/v5.x#toc_<fileName>` e.g. `https://deprecations.emberjs.com/v5.x#toc_ember-data-deprecate-legacy-imports` because file is `v5/ember-data-deprecate-legacy-imports.md`

kategengler:
I really don't think there is any harm in showing deprecations for future versions in the deprecation guides. After all, users could be using the beta or canary. Though the guides should be written as if the deprecation is already released. 

@runspired For URLs, since https://github.com/ember-learn/deprecation-app/pull/1261 a URL scheme of `https://deprecations.emberjs.com/id/<deprecation-id>` is available, and, I think, preferable. 

@Baltazore Thank you for PRing the deprecation guides.

content/ember-data/v5/ember-data-deprecate-computed-chains.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Deprecate Non Strict Types
+until: '6.0'
+since: '5.0'
+displayId: <none yet assigned>
+---
+
+This is a planned deprecation which will trigger when observer or computed chains are used to watch for changes on any EmberData RecordArray, ManyArray or PromiseManyArray.
```
kategengler, line 8: This deprecation should be written as if it has already happened and guess at the id.

Likely this deprecation also needs some before/after examples.

runspired: This one shouldn't be here. We implement a guard for a specific behavior to allow folks to opt out of legacy chains, we don't print a deprecation because we've never figured out how to detect legacy chains (the work you do to support them is eager, not lazy, so can't rely on "pull" to trigger).

Baltazore: removed this completely

content/ember-data/v5/ember-data-deprecate-relationship-remote-update-clearing-local-state.md
```diff
@@ -0,0 +1,83 @@
+---
+title: Deprecate Non Uniq Payloads
```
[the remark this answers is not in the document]
Baltazore: I do agree that without `Deprecate` it looks better and aligned with Ember and EmberCLI style, but we have `Deprecate` in most of v4 deprecations, should I change titles there as well ?

jaredgalanis (approved):

jaredgalanis:
merging since we're all released now

PR #1377, opened 2024-03-05
Deprecation guide for (action) / {{action}}

NullVoxPopuli:
Implementation PR: https://github.com/emberjs/ember.js/pull/20653
RFC: https://github.com/emberjs/rfcs/pull/1006

### DIFF
--- a/content/ember/v5/deprecate-template-action.md
+++ b/content/ember/v5/deprecate-template-action.md
@@ -1,7 +1,7 @@
 ---
 title: Action helper and modifier 
 until: 6.0.0
-since: 5.10.0
+since: 5.9.0
 ---
 
 ### Scenario: `action` is passed a string
--- a/content/ember/v5/deprecate-template-action.md
+++ b/content/ember/v5/deprecate-template-action.md
@@ -41,7 +41,22 @@ export default class Demo extends Component {
     }
 }
 ```
+After:
+```javascript
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+
+export default class Demo extends Component {
+    @action
+    plusOne() {
+       /* ... */ 
+    }
+}
+```
+
 or
+
+Before:
 ```javascript
 import Component from '@ember/component';
 
@@ -53,7 +68,22 @@ export default class Demo extends Component {
     }
 }
 ```
+
+After:
+```javascript
+import Component from '@ember/component';
+import { action } from '@ember/object';
+
+export default class Demo extends Component {
+    @action
+    plusOne() {
+       /* ... */ 
+    }
+}
+```
 or
+
+Before:
 ```javascript
 import Component from '@ember/component';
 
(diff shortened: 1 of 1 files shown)


content/ember/v5/deprecate-template-action.md
```diff
@@ -0,0 +1,138 @@
+---
+title: Action helper and modifier 
+until: 6.0.0
+since: 5.10.0
```
IgnaceMaes, line 4: Just double checking, is this deprecation targeting 5.10.0? (so not the next version)

content/ember/v5/deprecate-template-action.md
```diff
@@ -0,0 +1,138 @@
...
+</button>
+```
+or, if `plusOne` is passed in as an argument 
+```handlebars
+<button type="button" {{on 'click' @plusOne}}>
+  Click Me
+</button>
+```
+
+If the `plusOne` action is in an actions object, it needs to move out:
```
IgnaceMaes, line 30: Is it also possible that the action isn't defined in the direct parent component, as it bubbled up through the DOM tree? In that case the function reference will need to be passed as a component parameter. Would be good to also mention here.

It's been a years since I've used this syntax so could be mistaken though.

content/ember/v5/deprecate-template-action.md
NullVoxPopuli: depends on if we merge the implementation before 5.9 becomes beta?

content/ember/v5/deprecate-template-action.md
NullVoxPopuli: I think you're right -- I've updated the text to mention this.

content/ember/v5/deprecate-template-action.md
IgnaceMaes: Alright, just wanted to double check this. Sounds good then!

NullVoxPopuli:
I've updated the `since` version to 5.9

IgnaceMaes (approved):

content/ember/v5/deprecate-template-action.md
```diff
@@ -0,0 +1,170 @@
...
+import { action } from '@ember/object';
+
+export default Component.extend({
+  plusOne: action(function() {
+      /* ... */ 
+  }),
+})
+```
+
+If `(action)` or `{{action}}` is passed a string, it's _possible_ that the referenced method is declared on the caller, and _not_ the immediate component -- that is, `(action)` and `{{action}}` bubble up the render tree.
```
achambers, line 111: > that is, `(action)` and `{{action}}` bubble up the render tree.

This is only the case in route templates/controllers and is not the case for components.

content/ember/v5/deprecate-template-action.md
```diff
@@ -0,0 +1,170 @@
...
+After:
+```javascript
+import Component from '@ember/component';
+import { action } from '@ember/object';
+
+export default Component.extend({
+  plusOne: action(function() {
+      /* ... */ 
+  }),
+})
```
achambers, line 108: We should also capture the fact someone might be calling `this.send('plusOne');` which needs to change to `this.plusOne()`;

content/ember/v5/deprecate-template-action.md
```diff
@@ -0,0 +1,186 @@
...
+```
+
+If `(action)` or `{{action}}` is passed a string, it's _possible_ that the referenced method is declared on the caller, and _not_ the immediate component -- that is, `(action)` and `{{action}}` bubble up the render tree from route templates -> controllers -> routes.
+
+Note that `@action` is completely different from `(action)` or `{{action}}` (and is partly a motivator for deprecating `(action)` and `{{action}}`, to reduce ambiguity).
+
+`@action` is binds the `this` on the method to the instance of the class.
```
lukasnys, line 131:
```suggestion
`@action` binds the `this` on the method to the instance of the class.
```

PR #1380, opened 2024-03-17
chore: replace all filename comments in code examples by proper `data-filename`

IgnaceMaes:
We have support for file names in our code block showdown plugin, so let's use it consistently instead of using code comments.


MinThaMie (approved):
Love it :) Thanks for doing the work!

PR #1381, opened 2024-05-02
Deprecation guides for deprecating importing from `'ember''

NullVoxPopuli:
RFC: https://github.com/emberjs/rfcs/pull/1003
Implementation: https://github.com/emberjs/ember.js/pull/20686
And: https://github.com/emberjs/ember.js/pull/20813

Advancement: https://github.com/emberjs/rfcs/pull/1015

Until: 7.0.0
Since: 6.5.0

PR #1383, opened 2024-06-04
Add parent relationship to deprecations to allow grouping in the UI, especially in the index

kategengler:
- Added `since` to the metadata display since I felt it was missing especially on the id route
- Made id in the metadata display link to the individual id page to expose those URLs
- Added some info to the README about the new relationship but also about adding deprecation guides

No deprecations currently use this, though I did certainly test it out. I am aiming to use this for #1381 so we do not lose the usefulness of the index. Due to the way the `Deprecate `import Ember from 'ember'`` deprecation is done, we need a URL for every single API even if they weren't ones that were at all used by the public. I don't want the index to scare people off of upgrading to 6.0, but there should still be a way to see all the related deprecations.

cc @NullVoxPopuli 

CleanShot 2025-04-04 at 17 14 03

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,9 +5,26 @@
 
 This is the app that serves https://deprecations.emberjs.com/
 
+## Linking to deprecations
+
+You can link to a specific deprecation by using the ID of the deprecation. For example, to link to the deprecation with the ID `my-old-api`, you can use the following URL:
+`https://deprecations.emberjs.com/id/my-old-api`. These URLs can be generated in advance of adding the deprecation guide, when the deprecation lands in the code.
+When adding a deprecation the filename should match the ID of the deprecation, or the `displayId` should be specified in the frontmatter.
+
 ## Adding new deprecations
 
 The [content](https://github.com/ember-learn/deprecation-app/tree/main/content/) folder contains all the deprecations that are listed by the website. To add a new deprecation, add it to the appropriate folder by creating a new file. The content of the file needs to follow a specific format for the app to work. You can see [this sample](https://raw.githubusercontent.com/ember-learn/deprecation-app/main/content/ember/v3/getting-the-each-property.md) for reference.
+
+### Frontmatter
+
+#### Grouping deprecations
+
+```markdown
+  parent: deprecation-id
+```
+
+Can be used to nest deprecations under a parent grouping for the purpose of the UI. The deprecations will still be available via the direct ID URLs.
+
 
 ## Prerequisites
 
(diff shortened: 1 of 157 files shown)


NullVoxPopuli (approved):
behavior seems good :tada:

MinThaMie:
We discussed this during our learning meeting on 17th of June, and while we understand that it's a ton of deprecations we believe introducing this new data pattern/structure is not necessary and makes it more complicated to maintain. 

We have plans to structurally address the issue you ran into. I hope you understand :) Thanks for the PoC!

kategengler:
Do you have a timeline for addressing the issue? We're attempting to land the deprecation of `import Ember from 'ember'` which is the one that introduces 150 deprecations, most of which will not be relevant for most users.

MinThaMie:
We're happy to land that deprecation as is :) it does not seem to effect the usability of the page, so you could merge `import Ember from ember` right now if you want to.

kategengler:
Can we reconsider this? See the deprecations in 6.x to see the unusability https://deprecations.emberjs.com/v6.x

NullVoxPopuli:
this direct link sure is nice! 

https://deploy-preview-1383--ember-deprecations.netlify.app/id/deprecate-import--array-from-ember

NullVoxPopuli:
oh -- those links already existed. I had no idea.

The links from the main page are nice

kategengler:
@ember-learn/learning-core-team I updated this one and added it to `import Ember from 'ember'` to add an example now that it has landed.

NullVoxPopuli:
Unfortunately, because this is no longer SSR'd, all the deprecations have a significant render time on lower powered hardware:
I would not feel great about this (as is) until we get some rendering improvements landed in the rendering side of things (VM/etc).

Tho, we could have the best of both worlds if we still render everything to the page via `<details>`, which would mean that ctrl+f would still work, ya?

NullVoxPopuli:
> Having them all load on the page for every visitor is preferable, then?

relying on SSR to have them rendered (in some way, be that hidden (what you're proposing) or visible (what we do today) would solve the problem.

There is a bigger problem that is def out side the scope of this PR where the VM's DOM rendering is just too slow (we have to kill a bunch of tech debt in there to speed it up... maybe even get rid of the VM entirely, idk yet) but even in the fastest frontend JS framework out there right now, this amount of rendering would have noticeable delay if we wrap all the content in an `if`

kategengler:
I'm again wishing we had this so that there would be a landing page we could link to explain the overall deprecation of the barrel file.

PR #1384, opened 2024-06-06
make the deprecation anchor § link directly to the id route

mansona:
This is to encourage people to share direct links instead of sharing the strange anchor toc links that we had before 👍


jaredgalanis (approved):
👍

PR #1389, opened 2024-07-23
Mobile list styling

### LINKED ISSUE #1239: [BUG] Mobile  Deprecation version section is too long

MinThaMie:
To fix #1239 . I decided for a simpler approach than #1240 
Screenshot 2024-07-23 at 11 55 52

closes: https://github.com/ember-learn/deprecation-app/pull/1240

### DIFF
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -586,7 +586,9 @@ p a:hover {
         .ember-version-graphic {
           $height: 132px;
           width: $height;
-          height: $height;
+          min-height: $height;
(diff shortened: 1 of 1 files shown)


mansona (approved):
LGTM

PR #1398, opened 2025-03-23
fix: width issue when no code block

MehulKChaudhari:
Before: 
image

after: 
image

### DIFF
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -207,6 +207,18 @@ table {
   }
 }
 
+@media (min-width: 1007px) and (max-width: 1261px) {
+  .layout-grid {
+    grid-template:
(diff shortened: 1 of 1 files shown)


mansona (changes requested):
Alas this isn't the right solution 😞 
Screenshot 2025-04-02 at 09 28 32

MehulKChaudhari:
> Alas this isn't the right solution 😞 Screenshot 2025-04-02 at 09 28 32


I just checked and found that the content overflow issue (as shown in your screenshot) occurred in the main branch too, when width is between 1261px and 1007px.

The changes I have made now should fix this issue.
Thanks

kategengler (approved):

PR #1399, opened 2025-04-04
Update importing-inject-from-ember-service to have guidance for library authors that need to support ember prior to 4.1

NullVoxPopuli:


kategengler (approved):

IgnaceMaes (approved):

IgnaceMaes:
Merged with bypassing Percy, as it seems to be stuck.

PR #1401, opened 2025-04-07
enforce the right type for since and until in frontmatter

mansona:


kategengler:
Ah, I think the types were wrong only because it was only two digits `6.0` vs `6.0.0`.

PR #1402, opened 2025-04-08
Add RFCs to ember-data 5.x deprecations (for which I could find the RFCs)

kategengler:


content/ember-data/v5/ember-data-deprecate-store-extends-ember-object.md
```diff
@@ -28,4 +29,4 @@ If you are unsure whether your Store class uses EmberObject APIs, set this
 config and uses of those APIs will throw exceptions. The most common API that 
 may have been used is `Store.extend({...`.
 
-This deprecation is from RFC [#1026](https://rfcs.emberjs.com/id/1026-ember-data-deprecate-store-extends-ember-object).
+This deprecation was introduced in RFC [#1026](https://rfcs.emberjs.com/id/1026-ember-data-deprecate-store-extends-ember-object).
```
kategengler, line 32: I think we should consider adding this bit to the template in the future, conditional on the rfc field existing.

NullVoxPopuli (approved):

PR #1404, opened 2025-06-20
Add deprecation guide for Ember.Evented and @ember/object/events

wagenet:
RFC: https://github.com/emberjs/rfcs/pull/1111

### DIFF
--- a/content/ember/v6/evented.md
+++ b/content/ember/v6/evented.md
@@ -6,13 +6,13 @@ since: 6.6.0
 
 The `Ember.Evented` mixin, the underlying `@ember/object/events` module (`addListener`, `removeListener`, `sendEvent`), and the `on()` function from `@ember/object/evented` are all deprecated.
 
-These APIs provided a way for Ember objects to send and receive events. With modern JavaScript features and patterns, we recommend more explicit and standard approaches. For eventing, we recommend refactoring to use a library like [emittery](https://www.npmjs.com/package/emittery).
+These APIs provided a way for Ember objects to send and receive events. With modern JavaScript features and patterns, we recommend more explicit and standard approaches. For eventing, we recommend refactoring to use a modern asynchronous library like [emittery](https://www.npmjs.com/package/emittery) or (if you need to preserve synchronous semantics) a library such as [`nanoevents`](https://www.npmjs.com/package/nanoevents) or [`mitt`](https://www.npmjs.com/package/mitt).
+
+> ⚠️ Important: `Ember.Evented` emits events *synchronously*. Changing to a library with asynchronous behavior, while recommended, may lead to subtle changes in your application's behavior.
 
 Please note: The methods from `Evented` (`on`, `one`, `off`, `trigger`, `has`) were also available on `Ember.Component`, `Ember.Route`, and `Ember.Router`. While usage on these objects is deprecated, the methods will continue to be supported and not deprecated on the `RouterService`, since key parts of its functionality are difficult to reproduce without them.
 
 ### Replacing `Evented` with `emittery`
-
-We recommend the [emittery](https://www.npmjs.com/package/emittery) library, which is a modern, promise-based event emitter. This is useful for services that need to broadcast state changes across an application, such as a session service that announces login and logout events.
 
 First, add `emittery` to your project:
 ```bash


content/ember/v6/evented.md
```diff
@@ -0,0 +1,138 @@
...
+title: Ember.Evented and @ember/object/events
+until: 7.0.0
+since: 6.6.0
+---
+
+The `Ember.Evented` mixin, the underlying `@ember/object/events` module (`addListener`, `removeListener`, `sendEvent`), and the `on()` function from `@ember/object/evented` are all deprecated.
+
+These APIs provided a way for Ember objects to send and receive events. With modern JavaScript features and patterns, we recommend more explicit and standard approaches. For eventing, we recommend refactoring to use a library like [emittery](https://www.npmjs.com/package/emittery).
+
+Please note: The methods from `Evented` (`on`, `one`, `off`, `trigger`, `has`) were also available on `Ember.Component`, `Ember.Route`, and `Ember.Router`. While usage on these objects is deprecated, the methods will continue to be supported on the `Ember.RouterService`, since key parts of its functionality are difficult to reproduce without them.
```
NullVoxPopuli, line 11:
```suggestion
Please note: The methods from `Evented` (`on`, `one`, `off`, `trigger`, `has`) were also available on `Ember.Component`, `Ember.Route`, and `Ember.Router`. While usage on these objects is deprecated, the methods will continue to be supported and not deprecated on the `RouterService`, since key parts of its functionality are difficult to reproduce without them.
```

content/ember/v6/evented.md
```diff
@@ -0,0 +1,138 @@
...
+  handleLogin(user) {
+    console.log(`User logged in: ${user.name}`);
+    // ... update component state
+  }
+}
+```
+
+After refactoring to use `emittery`, the service manages its own event emitter and provides clear methods for subscribing.
+
+After:
```
NullVoxPopuli, line 78:
```suggestion
#### After
```

content/ember/v6/evented.md
```diff
@@ -0,0 +1,138 @@
...
+    this.#emitter.emit('loggedIn', userData);
+  }
+
+  logout() {
+    const oldUser = this.user;
+    this.user = null;
+    this.#emitter.emit('loggedOut', oldUser);
+  }
+
+  // Public subscription methods
```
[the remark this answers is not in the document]
NullVoxPopuli: but what you have is easier to be type-safe

content/ember/v6/evented.md
```diff
@@ -0,0 +1,138 @@
...
+since: 6.6.0
+---
+
+The `Ember.Evented` mixin, the underlying `@ember/object/events` module (`addListener`, `removeListener`, `sendEvent`), and the `on()` function from `@ember/object/evented` are all deprecated.
+
+These APIs provided a way for Ember objects to send and receive events. With modern JavaScript features and patterns, we recommend more explicit and standard approaches. For eventing, we recommend refactoring to use a library like [emittery](https://www.npmjs.com/package/emittery).
+
+Please note: The methods from `Evented` (`on`, `one`, `off`, `trigger`, `has`) were also available on `Ember.Component`, `Ember.Route`, and `Ember.Router`. While usage on these objects is deprecated, the methods will continue to be supported and not deprecated on the `RouterService`, since key parts of its functionality are difficult to reproduce without them.
+
+### Replacing `Evented` with `emittery`
```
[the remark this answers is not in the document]
bendemboski: Observers are async by default, `@ember/object/events` and `Evented` are synchronous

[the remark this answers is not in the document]
wagenet: I've updated the guides to make this clear.

content/ember/v6/evented.md
```diff
@@ -0,0 +1,138 @@
+---
+title: Ember.Evented and @ember/object/events
+until: 7.0.0
+since: 6.6.0
```
kategengler:
```suggestion
since: 7.4.0
```

content/ember/v6/evented.md
```diff
@@ -0,0 +1,138 @@
+---
+title: Ember.Evented and @ember/object/events
+until: 7.0.0
```
kategengler:
```suggestion
until: 8.0.0
```

content/ember/v7/evented.md
```diff
@@ -0,0 +1,138 @@
...
+
+These APIs provided a way for Ember objects to send and receive events. With modern JavaScript features and patterns, we recommend more explicit and standard approaches. For eventing, we recommend refactoring to use a modern asynchronous library like [emittery](https://www.npmjs.com/package/emittery) or (if you need to preserve synchronous semantics) a library such as [`nanoevents`](https://www.npmjs.com/package/nanoevents) or [`mitt`](https://www.npmjs.com/package/mitt).
+
+> ⚠️ Important: `Ember.Evented` emits events *synchronously*. Changing to a library with asynchronous behavior, while recommended, may lead to subtle changes in your application's behavior.
+
+Please note: The methods from `Evented` (`on`, `one`, `off`, `trigger`, `has`) were also available on `Ember.Component`, `Ember.Route`, and `Ember.Router`. While usage on these objects is deprecated, the methods will continue to be supported and not deprecated on the `RouterService`, since key parts of its functionality are difficult to reproduce without them.
+
```
kategengler, line 18: Shouldn't our first suggestion be to call methods directly where possible? If someone was using Evented with the example below I do not think my first suggestion would be to use Emittery but to refactor away from the pattern entirely.

I also think this needs examples for where it was used within Ember.Component, and Ember.Route and Router for the modern equivalents. 

For example, for Ember Component, we should suggest moving to Glimmer components but also mention that instead of  `on('didInsertElement')` that the hooks can be used directly. 

They were also common with ember-simple-auth so we should suggest upgrading that addon or moving to the newer patterns there.

content/ember/v7/evented.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: Ember.Evented and @ember/object/events
```
kategengler, line 2: Noticing now that this file needs to be moved to v7

NullVoxPopuli: done

content/ember/v7/evented.md
```diff
@@ -0,0 +1,312 @@
...
+    this.greeting = `Welcome back, ${user.name}`;
+  }
+
+  handleLogout() {
+    this.greeting = 'Please sign in';
+  }
+}
+```
+
+The service keeps the tracked state. The component derives from it:
```
kategengler:
```suggestion
The service owns the tracked state. The component derives from it:
```

content/ember/v7/evented.md
```diff
@@ -0,0 +1,312 @@
...
+
+  get greeting() {
+    let { user } = this.session;
+
+    return user ? `Welcome back, ${user.name}` : 'Please sign in';
+  }
+}
+```
+
+The component no longer keeps its own copy of the state, so the subscription and its destructor are gone. The two handlers that had to agree with each other are now one getter.
```
kategengler:
```suggestion
```

content/ember/v7/evented.md
```diff
@@ -0,0 +1,312 @@
...
+
+The same rewrite applies to state that is computed from the event payload. A component that listened for `recordsLoaded` to recount a list should read the list and count it in a getter:
+
+```javascript
+get completedCount() {
+  return this.todos.items.filter((item) => item.isComplete).length;
+}
+```
+
+Derived state cannot go stale, and there is no listener left to leak.
```
kategengler:
```suggestion
```

content/ember/v7/evented.md
```diff
@@ -0,0 +1,312 @@
...
+
+### Call methods directly
+
+Sometimes an event does trigger real work rather than a state update, such as clearing a cache or starting a request. If the object that triggers the event already knows the object that handles it, replace the event with a method call.
```
kategengler:
```suggestion
Sometimes an event does trigger real work rather than a state update, such as clearing a cache or starting a request. If the object that triggers the event already knows the object that handles it, replace the event with a method call:
```

content/ember/v7/evented.md
```diff
@@ -0,0 +1,436 @@
...
+
+// app/services/cart.js
+export default class CartService extends Service {
+  clear() { /* ... */ }
+}
+```
+
+The call is type-safe, and `logout` now states what it does instead of hiding the wiring in a subscription registered somewhere else.
+
+### Classic component lifecycle hooks
```
NullVoxPopuli, line 185: Evented and `@ember/component` are about to be deprecated, so this prose should be take that in to consideration.

The before examples should at least look like code that someone might accidentally write today, rather than inherited from 10 years ago

(a migration this old could be useful as a last example, I suppose)

PR #1405, opened 2025-06-20
Add deprecation guides for ArrayProxy, ObjectProxy, and PromiseProxyMixin

wagenet:
RFC: https://github.com/emberjs/rfcs/pull/1112

### DIFF
--- a/content/ember/v6/deprecate-array-proxy.md
+++ b/content/ember/v6/deprecate-array-proxy.md
@@ -1,0 +1,208 @@
+---
+title: Deprecation of ArrayProxy
+until: 7.0.0
+since: 6.5.0
+---
+
+`ArrayProxy` is deprecated. In modern Ember, you can replace `ArrayProxy` with native JavaScript classes and proxies. The best replacement depends on how you were using `ArrayProxy`. Some example use cases are shown below.
+
+### Swapping Content
+
+If you were using `ArrayProxy` to easily swap out the underlying array while keeping a stable reference, you can achieve a similar, transparent effect using a native `Proxy` backed by a class with a `@tracked` property.
+
+Before:
+
+```javascript
+import ArrayProxy from '@ember/array/proxy';
+import { A } from '@ember/array';
+
+const pets = A(['dog', 'cat', 'fish']);
+const proxy = ArrayProxy.create({ content: pets });
+
+proxy.get('firstObject'); // 'dog'
+
+// Later, you can easily swap the content
+proxy.set('content', A(['amoeba', 'paramecium']));
+
+proxy.get('firstObject'); // 'amoeba'
+```
+
+After:
+
+```javascript
+import { tracked } from '@glimmer/tracking';
+
+// A helper class to hold the tracked state.
+class SwappableState {
+  @tracked content;
+
+  constructor(initialContent) {
+    this.content = initialContent;
+  }
+}
+
+// A factory function to create a proxy that is transparent
+// and allows swapping the underlying content.
+function createSwappableArray(initialContent) {
+  const state = new SwappableState(initialContent);
+
+  return new Proxy(state, {
+  get(target, property, receiver) {
+    // Allow getting/setting the content directly for swapping.
+    if (property === 'content') {
+      return target.content;
+    }
+    // Delegate all other property access to the content array.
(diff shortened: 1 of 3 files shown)


content/ember/v6/deprecate-array-proxy.md
```diff
@@ -0,0 +1,208 @@
+---
+title: Deprecation of ArrayProxy
+until: 7.0.0
+since: 6.5.0
+---
+
+`ArrayProxy` is deprecated. In modern Ember, you can replace `ArrayProxy` with native JavaScript classes and proxies. The best replacement depends on how you were using `ArrayProxy`. Some example use cases are shown below.
+
+### Swapping Content
```
kategengler, line 9: These guides show a way to achieve something technically the same as what people had before -- which is great -- but I also think we should guide users into what the modern Ember pattern would do instead. In this case, I believe that would mostly be tracked-built-ins.

content/ember/v6/deprecate-promise-proxy-mixin.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+And in your template:
+
+```handlebars
+{{#let this.loadData.last as |taskInstance|}}
+  {{#if taskInstance.isRunning}}
+    Loading...
+  {{else if taskInstance.isSuccessful}}
+    Value: {{taskInstance.value.value}}
```
NullVoxPopuli, line 65:
```suggestion
    Value: {{taskInstance.value}}
```

content/ember/v6/deprecate-promise-proxy-mixin.md
```diff
@@ -0,0 +1,114 @@
...
+{{#if this.isLoading}}
+  Loading...
+{{else if this.content}}
+  Value: {{this.content.value}}
+{{else if this.error}}
+  Error: {{this.error}}
+{{/if}}
+```
+
+Using a library like [ember-concurrency](https://ember-concurrency.com/docs/introduction) is highly recommended for managing asynchronous tasks in Ember applications, as it provides robust solutions for handling loading/error states, cancellation, and more.
```
NullVoxPopuli, line 114:
```suggestion
Using a library like [ember-concurrency](https://ember-concurrency.com/docs/introduction) is highly recommended for managing concurrent asynchronous user-initiated tasks in Ember applications, as it provides robust solutions for handling loading/error states, cancellation, and more.
```

content/ember/v6/deprecate-array-proxy.md
kategengler: My general point is that we should point people to "the Ember way" going forwards, not just a strict technical replacement -- why did they want a swappable container? Do you still need that with @tracked? (and applying to the other guides not just this first one that I commented on).

ef4:
I suspect the biggest hazard when trying to clear this deprecation is that computeds elsewhere in your app that depend on an ArrayProxy aren't going to react the same when you switch to TrackedArray. Maybe we should make a recommendation about removing computeds first before tackling this one?

wagenet:
I _also_ want to deprecate computeds 😆

content/ember/v6/deprecate-array-proxy.md
```diff
@@ -0,0 +1,273 @@
...
+
+proxy.get('firstObject'); // 'dog'
+
+// Later, you can easily swap the content
+proxy.set('content', A(['amoeba', 'paramecium']));
+
+proxy.get('firstObject'); // 'amoeba'
+```
+
+After:
```
NullVoxPopuli, line 55: I think it would be good to add two approaches for "After" in `<details>` for:
- exact technical replacement (change no consuming code) (what is shown here already)
- recommended migration path for canonical array using

wagenet:
I've added a note about ComputedProperties.

content/ember/v6/deprecate-array-proxy.md
wagenet: The description on this section reads:

> If you were using `ArrayProxy` to easily swap out the underlying array while keeping a stable reference, you can achieve a similar, transparent effect using a native `Proxy` backed by a class with a `@tracked` property. Again, prefer tracked arrays and properties if you can refactor your code to use them.

So I think it's clear that this is the "exact technical" replacement.

PR #1406, opened 2025-06-20
Add Array deprecation guides

wagenet:
RFC: https://github.com/emberjs/rfcs/pull/1114

Note that Ember.ArrayProxy deprecation guides are in #1405.


wagenet:
I've added a note about Computed Properties.

PR #1407, opened 2025-06-20
Add deprecation guides for observers

wagenet:


wagenet:
Looks like .hbs code blocks aren't allowed. Is this intentional?

content/ember/v6/deprecate-ember-object-observable.md
```diff
@@ -0,0 +1,147 @@
...
+### A Note on Legacy Computed Properties and Setters
+
+When working with legacy computed properties, the way you set matters for reactivity.
+
+#### Updating Plain Properties
+
+To trigger reactivity (like re-computing a dependent computed property) when changing a plain property, you **must** use the `set` function. A native JavaScript assignment (`person.firstName = 'Jane'`) will change the value but will **not** trigger reactivity.
+
+```javascript
+import { computed, set } from '@ember/object';
```
ef4, line 90: I think instead of telling people to use `set` we should tell people to handle this case by refactoring their `@computed` away.

wagenet: Yeah, that's a good call

NullVoxPopuli: Depending on where folks are with their migration, it could be good to have both -- the migration we have here -- but also how we would do it today using `@tracked`.

We could even use `<details>` to allow people to choose their own adventure (and not overwhelm them with a skyscraper of code blocks)

content/ember/v6/deprecate-ember-object-observable.md
```diff
@@ -0,0 +1,147 @@
...
+  lastName = 'Doe';
+
+  @computed('firstName', 'lastName')
+  get fullName() {
+    return `${this.firstName} ${this.lastName}`;
+  }
+  set fullName(value) {
+    const [firstName, lastName] = value.split(' ');
+    // Note: `set` is still used inside the setter itself
+    set(this, 'firstName', firstName);
```
NullVoxPopuli, line 133: set is not imported 🙈

content/ember/v6/deprecate-ember-object-observers.md
```diff
@@ -0,0 +1,213 @@
+---
+title: Deprecation of @ember/object/observers
+until: 7.0.0
+since: 6.5.0
+---
+
+The `addObserver` and `removeObserver` methods from `@ember/object/observers` are deprecated. Instead of using observers, you should use tracked properties and native getters/setters.
```
NullVoxPopuli, line 7: we need another use case for before/after:
- ember-concurrency's `waitFor` uses observers

and whatever these do:
- ember-animated
- ember-data

ef4:
We discussed that this is another place where it's probably important to expose well-curated and suitably-low-level-looking API for doing integrations with the auto tracking system that are not just our own DOM renderer. That would enable cases like the ember-concurrency example to do efficient integration without polling, rAF etc.

PR #1408, opened 2025-06-20
Add Mixin deprecation guide

wagenet:
RFC: https://github.com/emberjs/rfcs/pull/1116

### DIFF
--- a/content/ember/v6/deprecate-ember-mixin.md
+++ b/content/ember/v6/deprecate-ember-mixin.md
@@ -4,7 +4,7 @@ until: 7.0.0
 since: 6.5.0
 ---
 
-`Ember.Mixin` is deprecated. This is part of the legacy `@ember/object` API and is not fully compatible with native classes. Instead of using mixins, you should refactor your code to use class-based patterns.
+`Ember.Mixin` is deprecated. This is part of the legacy `@ember/object` API and is not compatible with native classes. Instead of using mixins, you should refactor your code to use class-based patterns.
 
 One common pattern to replace mixins is to use class decorators.
 
@@ -71,4 +71,4 @@ export default class Comment extends EmberObject {
 }
 ```
 
-This approach provides the same functionality as the mixin but uses standard JavaScript features, making the code easier to understand and maintain.
+This approach provides the same functionality as the mixin but uses standard JavaScript features, making the code easier to understand.


content/ember/v6/deprecate-ember-mixin.md
```diff
@@ -0,0 +1,74 @@
+---
+title: Deprecating Ember.Mixin
+until: 7.0.0
+since: 6.5.0
+---
+
+`Ember.Mixin` is deprecated. This is part of the legacy `@ember/object` API and is not fully compatible with native classes. Instead of using mixins, you should refactor your code to use class-based patterns.
```
NullVoxPopuli, line 7:
```suggestion
`Ember.Mixin` is deprecated. This is part of the legacy `@ember/object` API and is not compatible with native classes. Instead of using mixins, you should refactor your code to use class-based patterns.
```

content/ember/v6/deprecate-ember-mixin.md
```diff
@@ -0,0 +1,74 @@
+---
+title: Deprecating Ember.Mixin
+until: 7.0.0
+since: 6.5.0
+---
+
+`Ember.Mixin` is deprecated. This is part of the legacy `@ember/object` API and is not compatible with native classes. Instead of using mixins, you should refactor your code to use class-based patterns.
```
kategengler, line 7: My strategy for removing mixins from a codebase has been to copy the code into the places that mixed in the mixin and then if there is any common state, extract that to a service. From there I refactor to utilities or to the service to clean it up.

NullVoxPopuli: See also: https://github.com/tc39/proposal-decorators?tab=readme-ov-file#classes

content/ember/v6/deprecate-ember-mixin.md
```diff
@@ -0,0 +1,74 @@
...
+
+    edit() {
+      console.log('starting to edit');
+      this.isEditing = true;
+    }
+  };
+}
+```
+
+Then, apply the decorator to your class:
```
ef4, line 61: Showing how to do this with a plain function call rather than decorator syntax might be good.

Especially because we still offer babel-style decorator call signatures, not the standards track stage3 ones.

NullVoxPopuli:
superseded by https://github.com/ember-learn/deprecation-app/pull/1434

NullVoxPopuli:
Closing in favor of #1434

PR #1412, opened 2025-08-12
Add deprecation guides for ComputedProperties

wagenet:


content/ember/v6/deprecate-import-alias-from-object-computed.md
```diff
@@ -0,0 +1,34 @@
...
+until: 7.0.0
+since: 6.0.0
+---
+
+`alias` from `@ember/object/computed` is deprecated.
+
+It created a second property that reflected another property's value.
+
+## Migration
+Prefer direct access to the original tracked state, or use a simple getter and setter.
```
kategengler, line 12: `alias` could point to a computed property, a getter, or really anything so this may need to describe how make sure they work going forward.

content/ember/v6/deprecate-import-and-from-object-computed.md
```diff
@@ -0,0 +1,30 @@
...
+import { and } from '@ember/object/computed';
+class Cart {
+  items = [];
+  user = null;
+  @and('items.length', 'user.isLoggedIn') canCheckout;
+}
+```
+After:
+```js
+class Cart {
```
kategengler, line 23: This needs reactivity

content/ember/v6/deprecate-import-computed-property-from-object-computed.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Migration
+Do not construct or refer to `ComputedProperty` directly. Use native getters with `@tracked` state instead.
+
+Before:
+```js
+import ComputedProperty from '@ember/object/computed';
+// rarely used directly
+```
+After: Remove the import; refactor any custom meta-programming to simple getters.
```
kategengler, line 17: I think custom ComputedProperties weren't rare.

content/ember/v6/deprecate-import-empty-from-object-computed.md
```diff
@@ -0,0 +1,28 @@
...
+```js
+import { empty } from '@ember/object/computed';
+class ListState {
+  items = [];
+  @empty('items') isEmpty;
+}
+```
+After:
+```js
+import { isEmpty } from '@ember/utils';
```
kategengler, line 22: `@ember/utils` has a deprecation RFC that has not landed https://github.com/emberjs/rfcs/blob/master/text/0334-deprecate-ember-utils.md?plain=1

ef4:
RFC review notes:

 - landing this deprecation is valuable because it teaches people to stop making new ones
 - this one might be hard to clear, so we would consider giving it a longer time before removal
 - before removal, we should strongly consider shipping some tools to help automate the migration 
 - we probably need to document what addons to upgrade before you attempt this (since dependency chains might bottom out down in addons)

PR #1427, opened 2026-02-15
Frontmatter validation

tcjr:
This PR updates the existing content frontmatter validation to require cleaner authoring data.

### Highlights
* **Improved, Stricter Authoring of Content**: We were already checking that `since` and `until` were string values. This PR extends the validation to require valid semver strings.
* **Data Cleanup**: This PR updates all existing invalid `since` and `until` values to be valid semver strings (`MAJOR.MINOR.PATCH`).
* **Lint, not Test**: Moves the frontmatter format validation from a _test_ to a _lint_.  It is now called using the `lint:frontmatter` script.
* **Better output**: The mocha test has been updated to say "_filename_ has valid frontmatter". The linter only shows failures to the console. (See screenshots below.)

#### Success
passing

#### Fail
failing


content/ember/v1/instance-initializers.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: Access to Instance Initializers
-until: ""
-since: ""
+until: 2.0.0
+since: 1.0.0
```
tcjr, line 4: This is a guess.  See #1428

kategengler: The `until` could be safely put at 2.0 for all the v1 deprecations

kategengler:
Do you think you could alter the left nav to only have major.minor? I expect that is why the data was mostly in there as "6.7", for example.

tcjr:
> Do you think you could alter the left nav to only have major.minor? I expect that is why the data was mostly in there as "6.7", for example.

Yes, definitely. I will handle this in the Ember app as a presentation concern. The data can remain a [valid semver value](https://semver.org/#spec-item-2) in the source yaml.

ember-cli-build.js
```diff
@@ -57,9 +57,5 @@ module.exports = function (defaults) {
     },
   });
 
-  app.import('node_modules/semver-compare/index.js', {
```
tcjr, line 60: A small side benefit of including the real `semver` package is having proper module imports, so this goes away.

tcjr:
> Do you think you could alter the left nav to only have major.minor? I expect that is why the data was mostly in there as "6.7", for example.

I made this change. It means that there are a few places in the visual regression tests that are failing because the data was inconsistent before -- [some places](https://deprecations.emberjs.com/v3.x) used `major.minor` and [some places](https://deprecations.emberjs.com/v4.x)  used `major.minor.patch`.

These changes causes it to look the same everywhere.

kategengler (approved):

PR #1429, opened 2026-02-21
GJS Components

tcjr:
This PR converts all the components in this app to single-file GJS components. This includes:

### 4 template-only conversions (hbs -> gjs):
- LayoutProvider
- MainLayout
- SideBarLayout
- TableOfContents

### 4 Glimmer component conversions (js+hbs -> gjs):
- EmberVersionGraphic
- DeprecationArticle
- SearchInput
- SearchResult
- SemVer

### 1 Ember component conversion (js+hbs -> gjs):
- TocSection


app/helpers/or.js
```diff
@@ -0,0 +1,7 @@
+import { helper } from '@ember/component/helper';
+
+function or([a, b]) {
```
IgnaceMaes, line 3: q: assuming these just worked without defining them before, as some dependency likely has ember-truth-helpers as subdependency?

IgnaceMaes (approved):

app/helpers/or.js
tcjr: They were magically resolved (likely from ember-truth-helpers via ember-styleguide). However, ember-truth-helpers is not a direct dependency of this app.  So, rather than add a new dependency pinned to an old version -- latest ember-truth-helpers doesn't support the old version of Ember this app uses -- I opted to create these two (`or` and `not`).

IgnaceMaes: Yeah makes sense

IgnaceMaes:
Thanks for the improvements! 🙌

PR #1431, opened 2026-05-05
Add package override examples for ember dependencies

NullVoxPopuli:
Added examples of package overrides for pnpm and npm to address build-time deprecation warnings for ember-cli-htmlbars and ember-cli-babel.


ef4:
As far as I know, there's no explained rationale for why everybody would need to force upgrade that ember-cli-babel version. We did a lot of compatibility work to keep the older ember-cli-babel's working the same in v1 addons.

PR #1432, opened 2026-06-04
docs(ember): add deprecation guide for Comparable mixin

olenderhub:
Adds the deprecation guide for `deprecate-comparable-mixin`.

This matches the deprecation entry added in Ember:

- `id: deprecate-comparable-mixin`
- `since: 7.2.0`
- `until: 8.0.0`
- URL: `https://deprecations.emberjs.com/id/deprecate-comparable-mixin`

The guide explains that apps and addons should stop importing/extending the `Comparable` mixin and instead define a function-valued `compare(other)` method directly.

Refs https://github.com/emberjs/ember.js/pull/21437


ef4 (approved):

content/ember/v7/deprecate-comparable-mixin.md
```diff
@@ -0,0 +1,37 @@
+---
+title: Deprecate Comparable mixin
+until: 8.0.0
+since: 7.2.0
```
kategengler, line 4: @NullVoxPopuli did this actually land in 7.2?

NullVoxPopuli: yeah: https://github.com/emberjs/ember.js/pull/21437

kategengler: It did, but it is also until 7.5 because it was private, so just an intimate api

content/ember/v7/deprecate-comparable-mixin.md
```diff
@@ -0,0 +1,37 @@
...
+---
+
+The `Comparable` mixin is deprecated.
```
kategengler:
```suggestion
The `Comparable` mixin is deprecated. This mixin was private but as it may have been intimate API, per our policy we've added a deprecation that will remain for at least one LTS. 
```

content/ember/v7/deprecate-comparable-mixin.md
NullVoxPopuli: updated

